Issue #41, opened 2022-09-09
disconnection is sometimes unrecoverable without restarting the app


mrluc:
I was testing what happens when closing down the server started with `iex -S phx.server` and restarting, and ran into some crashes, and more frequently, unrecoverable error messages. (Nothing but ❤️ for this project, when it's connected it's already magical, totally understand that it's [v0.0.1-alpha software](https://dockyard.com/blog/2022/09/09/liveview-native-weekly-update-1), totally fine if answer is 'WONTFIX, wait for next version' or something. I'm documenting this in case it's helpful to someone. I'm curious if the details I mention here are known current limitations, or if those who have used this a bit more know something I don't -- like, is this only an issue in dev but not really in practice? maybe due to livereload paths being hit? etc -- be cool to look at the elixirconf ios chat app to see how it's handled there 😄).

**Observed in:** a working 'Getting Started' tutorial app, at the point of list > detail navigation, but before registering a custom component, [ie at the end of this tutorial]( https://liveviewnative.github.io/liveview-client-swiftui/tutorials/phoenixliveviewnative/03-navigation-and-hero)). Edit: in the Simulator, on an M1 Air.

**Problem:** killing the Elixir server, and then restarting it, often results in the app being 'stuck' in a broken state, at least in my experience. Restarting the mobile app fixes it.

This can happen in a variety of ways:
- 👍 If I restart the Elixir server very quickly, within 1-2 seconds, all is usually well despite the momentary error.
- 👎 👍  If I restart the Elixir server somewhat quickly, approx 3-5 seconds, the app gets backgrounded! Kind of funny. When I un-background it, it shows the error screen initially and then recovers.
- ❌ however, from the root view, if I kill the server and restart in about 10 seconds or more, I usually encounter this most-frequent kind of fail, where the UI disagrees with logging! The error continues to be displayed, but both the server and the XCode console show that reconnection succeeded, and log heartbeat messages, etc; screenie showing this case:

  <img width="1000" alt="image" src="https://user-images.githubusercontent.com/38204/189452186-458ee705-fd14-43b6-bd09-f92a48bf5872.png">
- 👍 likewise, if I disconnect the server while in a child view, see the error, and then restart the server after 20 seconds or so, the child view does not recover -- but when I navigate back to the parent view, it will sometimes look good, and continue to function normally (perhaps due to navigation triggering an explicit reconnection?)
[...]

PR #51, opened 2022-09-16
attempt to fix 2 cases of UI not recovering after reconnection

mrluc:
Last week I ran into 2 specific cases where a simple tutorial app would, on reconnection if the phoenix server goes down for some reason, appear unable to recover to the end-user. See [the warnings highlighted in that screenshot from that issue](https://github.com/liveviewnative/liveview-client-swiftui/issues/41) for more info.

The errors were hinting very strongly that both were caused by a 'needs to be on the main thread, because it affects the UI' kind of thing, I checked to see if wrapping code that triggered those warnings in `DispatchQueue.main.async { ... }` would solve these (previously very reproducible for me) issues, and seems 👍 after doing some testing.

### DIFF
--- a/Sources/PhoenixLiveViewNative/LiveViewCoordinator.swift
+++ b/Sources/PhoenixLiveViewNative/LiveViewCoordinator.swift
@@ -343,7 +343,9 @@ public class LiveViewCoordinator<R: CustomRegistry>: ObservableObject {
             if case .awaitingSocketConnection(let continuation) = self.internalState {
                 continuation.resume(throwing: Error.socketError(error))
             } else {
-                self.internalState = .connectionFailed(.socketError(error))
+                DispatchQueue.main.async {
+                    self.internalState = .connectionFailed(.socketError(error))
+                }
             }
         }
         socket.logger = { message in logger.debug("[Socket] \(message)") }
@@ -479,8 +481,10 @@ public class LiveViewCoordinator<R: CustomRegistry>: ObservableObject {
         // todo: what should happen if decoding or parsing fails?
         self.rendered = try! Root(from: FragmentDecoder(data: renderedPayload))
         let elements = try! self.parseDOM(html: self.rendered.buildString(), baseURL: self.currentURL)
-        self.internalState = .connected
-        self.elements = elements
+        DispatchQueue.main.async {
+            self.internalState = .connected
+            self.elements = elements
+        }
     }
     
     private func handleLiveRedirect(_ payload: Payload) {
--- a/Sources/PhoenixLiveViewNative/LiveViewCoordinator.swift
+++ b/Sources/PhoenixLiveViewNative/LiveViewCoordinator.swift
@@ -343,7 +343,7 @@ public class LiveViewCoordinator<R: CustomRegistry>: ObservableObject {
             if case .awaitingSocketConnection(let continuation) = self.internalState {
                 continuation.resume(throwing: Error.socketError(error))
             } else {
-                DispatchQueue.main.async {
+                Task { @MainActor in
                     self.internalState = .connectionFailed(.socketError(error))
                 }
             }
@@ -419,7 +419,9 @@ public class LiveViewCoordinator<R: CustomRegistry>: ObservableObject {
             setupChannelJoinHandlers(channel: channel, token: token)
         }
         
-        handleJoinPayload(renderedPayload: renderedPayload)
+        Task { @MainActor in
+            handleJoinPayload(renderedPayload: renderedPayload)
+        }
     }
     
     private func setupChannelJoinHandlers(channel: Channel, token: ConnectionAttemptToken) {
@@ -440,7 +442,9 @@ public class LiveViewCoordinator<R: CustomRegistry>: ObservableObject {
                     continuation.resume(returning: renderedPayload)
                 } else if self.internalState.reconnectAutomatically {
                     // if we're in a state where, e.g., a previous attempt failed, but an automatic rejoin attempt succeeds, finish the join
-                    self.handleJoinPayload(renderedPayload: renderedPayload)
+                    Task { @MainActor in
+                        self.handleJoinPayload(renderedPayload: renderedPayload)
+                    }
                 } else {
                     // if we've received a message but don't have anything to do with it,
                     // assume that that will continue to be the case, and leave the channel to prevent future messages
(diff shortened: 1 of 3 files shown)


Sources/PhoenixLiveViewNative/Views/PhxModernNavigationLink.swift
```diff
@@ -68,10 +68,12 @@ struct PhxModernNavigationLink<R: CustomRegistry>: View {
...
-                navCoordinator.sourceRect = source?.frameProvider() ?? .zero
-                navCoordinator.sourceElement = source?.element
-                navCoordinator.overrideOverlayView = overrideView?.view
-                navCoordinator.navigationPath.append(dest)
+                DispatchQueue.main.async {
+                    navCoordinator.sourceRect = source?.frameProvider() ?? .zero
+                    navCoordinator.sourceElement = source?.element
+                    navCoordinator.overrideOverlayView = overrideView?.view
+                    navCoordinator.navigationPath.append(dest)
+                }
```
shadowfacts, line 76: This change is unnecessary. The warning from SwiftUI that you probably saw about changes being made during a view update is spurious (see the comment several lines below this).

Sources/PhoenixLiveViewNative/NavigationCoordinator.swift
```diff
@@ -16,7 +16,9 @@ class NavigationCoordinator: NSObject, UINavigationControllerDelegate, Observabl
             // while we're animating, don't update the currentRect otherwise it starts animating back
             // to the now-offscreen position of the source
             if case .idle = state {
-                currentRect = sourceRect
+                DispatchQueue.main.async {
+                    self.currentRect = self.sourceRect
+                }
```
shadowfacts, line 21: This is unnecessary. The `sourceRect` is only set from the button handler of `PhxModernNavigationLink`, which is always invoked on the main thread.

Sources/PhoenixLiveViewNative/LiveViewCoordinator.swift
```diff
@@ -479,8 +481,10 @@ public class LiveViewCoordinator<R: CustomRegistry>: ObservableObject {
         // todo: what should happen if decoding or parsing fails?
         self.rendered = try! Root(from: FragmentDecoder(data: renderedPayload))
         let elements = try! self.parseDOM(html: self.rendered.buildString(), baseURL: self.currentURL)
-        self.internalState = .connected
-        self.elements = elements
+        DispatchQueue.main.async {
+            self.internalState = .connected
+            self.elements = elements
+        }
```
shadowfacts, line 487: `handleJoinPayload` should always be called on the main thread, so this should be unnecessary. But, the call to it in `setupChannelJoinHandlers` (line 443) needs to be wrapped in a `Task { @MainActor in }`.

Sources/PhoenixLiveViewNative/LiveViewCoordinator.swift
```diff
@@ -343,7 +343,9 @@ public class LiveViewCoordinator<R: CustomRegistry>: ObservableObject {
             if case .awaitingSocketConnection(let continuation) = self.internalState {
                 continuation.resume(throwing: Error.socketError(error))
             } else {
-                self.internalState = .connectionFailed(.socketError(error))
+                DispatchQueue.main.async {
```
shadowfacts, line 346: Use the new Swift Concurrency style of switching actors, to match the rest of the file:

```suggestion
                Task { @MainActor in
```

shadowfacts (changes requested):
Thanks for the pull request! Some of these changes are a bit overzealous, but you're on the right track and I think this does resolve (part of) #41.

Sources/PhoenixLiveViewNative/LiveViewCoordinator.swift
mrluc: Oh, yeah. New Swift Concurrency. I know all about that, for sure. Haven't been living under a rock at all.

After all, people have paid me for projects I've written in Swift in the past, so I'm the Sort Of Person Who Probably Knows About New Swift Concurrency.

/me googles for 3 minutes

Seems nice.

So, those changes I did can probably be replaced by just this, and that same thing on line 443. Keep everything on the main thread in the first place, which we can afford to do, and rm the rest of the changes 👍. Will test that out.

shadowfacts:
FYI, (I haven't checked but) it sounds like the spurious warning in the navigation link is fixed in the most recent Xcode 14.1 beta: https://www.donnywals.com/xcode-14-publishing-changes-from-within-view-updates-is-not-allowed-this-will-cause-undefined-behavior/

mrluc:
@shadowfacts interesting, I'm only on 14.0.1, I'll check before/after the beta when I circle back to this - I imagine there could be some cases of bugs like that; last Friday I read a bit about `@MainActor`, and I see that Swift UI's built-in components _should_ have all been annotated with that, so that all of their publishing should happen on MainActor automatically. Which would lean towards that chap having hit a legitimate bug.

That learning **did** make me wonder if the best way to do this now might be to annotate some of the LVN classes, those with observable properties used in UI, like `NavigationCoordinator`. I ran across that in [this blog post](https://www.swiftbysundell.com/articles/the-main-actor-attribute/#custom-ui-related-classes); supposedly this kind of annotation will automagically ensure that all publishing happens on main thread:

```swift
@MainActor class NavigationCoordinator: NSObject, UINavigationControllerDelegate, ObservableObject {
 // ^^ add this
    // ...
    @Published var overrideOverlayView: AnyView?
    @Published var sourceElement: Element?
    // ...
}
```

Ofc we can address it either way -- at the 'source', or downstream at the 'sink'; this annotation is at the 'sink', at publish time. Kicking off the async actions on mainactor, as you mentioned, would be going to the 'source' -- but now that I see they have this more automatic sugar for publishing back to main, it seems like this might be the preferred way?

Disclaimer: this is conjecture, I'm not able to work on this today, hehe. I've scheduled a 'shop day' later this week, so I'll probably give this all a a try then if it's still relevant.

Sorry for the delay -- I blame it entirely on this beautiful Minnesotan autumn, which has stolen all of my 'shop time' these past few weeks 😊 🍂.

mrluc:
Aight, had time to run some tests, and I can confirm that those 2 suggested more-upstream `Task { @MainActor in ... }` blocks keep the same improvements as the changes in this PR 👍.

I tested and kept notes on what works/fails in each case, currently looks like this.

| server unavailable: | for: | reconnects in: | result (master) | result (this PR) |
|-|-|-|-|-|
|initial connection | 0-10s | (never reconnects) | unrecoverable UI | unrecoverable UI |
|viewing root liveview| 0-10s | root view | often a fatalError | ✅  fixed |
|viewing child view | 0-10s | child view | often a fatalError | ✅ fixed |
|viewing child view | 0-10s | root view | fatalError/unrecoverable UI | unrecoverable UI |

vid showing the 3 last cases above, before and after (not showing initial connection case):

| before | after |
|-|-|
| <video src="https://user-images.githubusercontent.com/38204/192702783-ae43b27e-8488-43da-8cb1-ae1782abdc98.mp4" /> | <video src="https://user-images.githubusercontent.com/38204/192806144-0780f6a5-1c82-448b-8507-0c0f8cbc2899.mp4" /> |

**Next steps:** those 2 cases this PR doesn't touch may be attractive as bite-sized issues

1. frozen UI if initial connection attempt doesn't succeed
2. frozen UI if navigating back during disconnection

shadowfacts (approved):

shadowfacts:
Thanks for contributing! I'll look into the other two cases.

PR #130, opened 2023-01-10
Split `LiveViewCoordinator` into Session Coordinator and View Coordinator

carson-katri:
This splits the `LiveViewCoordinator` into a `LiveSessionCoordinator` and `LiveViewCoordinator`. The session coordinator handles a single global websocket. The view coordinator handles the Phoenix channels for each connected LiveView. This allows multiple LiveViews to be connected at a time. This is primarily to allow for things like split screen navigation on iPad, bottom sheets on iOS, and other scenarios where multiple screens need to be interactive at the same time.

Take a look and let me know what you think. Is this a direction we should go in?

It currently doesn't work with `push_redirect` as that ends the connection to the previous LiveView. Instead you need to send the `native_redirect` event:
```ex
def push_native_redirect(socket, opts \\ []) do
  push_event(
    socket,
    "native_redirect",
    %{
      to: Keyword.get(opts, :to),
      kind: if(Keyword.get(opts, :replace, false), do: :replace, else: :push),
      mode: :multiplex
    }
  )
end
```
Let me know if there is a better solution for this.


Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
```diff
@@ -0,0 +1,332 @@
...
+        }
+        let diff = try RootDiff(from: FragmentDecoder(data: payload))
+        self.rendered = try self.rendered.merge(with: diff)
+        self.document?.merge(with: try Document.parse(self.rendered.buildString()))
+    }
+    
+    private nonisolated func parseDOM(html: String, baseURL: URL) throws -> Elements {
+        // we parse the DOM as XML rather than HTML, because we don't want it to apply browser-specific HTML transformations
+        // such as <image> -> <img>
+        let document = try Parser.xmlParser().parseInput(html, baseURL.absoluteString)
```
shadowfacts, line 130: This can actually be changed back to the HTML parser, since that's all it's used for, now that core handles parsing the live view.

Sources/LiveViewNative/LiveView.swift
```diff
@@ -24,114 +22,100 @@ import LVNObjC
...
-                            .animation(navigationCoordinator.state.animation, value: navigationCoordinator.currentRect)
-                    }
-                    .edgesIgnoringSafeArea(.all)
-                }
+        if case .enabled = session.config.navigationMode {
+            if UIDevice.current.userInterfaceIdiom == .pad {
+                navigationSplitView
+            } else {
+                navigationStack
             }
```
shadowfacts, line 77: I don't think using a split view should be automatic. The navigation stack matches the behavior of regular LiveView on the web and the split view is a far enough departure that it should be opt-in.

Sources/LiveViewNative/NavStackEntryView.swift
```diff
@@ -10,15 +10,11 @@ import LiveViewNativeCore
 
 struct NavStackEntryView<R: CustomRegistry>: View {
     // this is a @StateObject instead of @ObservedObject because changing which DOMWindows for a LiveView is not allowed
```
shadowfacts, line 12:
```suggestion
```

shadowfacts (changes requested):
I've requested some minor changes, but I think this is definitely the right approach. Good work

Sources/LiveViewNative/BuiltinRegistry.swift
```diff
@@ -57,9 +58,9 @@ struct BuiltinRegistry {
     
     enum ModifierType: String {
         case frame
-        case listRowInsets = "list_row_insets"
-        case listRowSeparator = "list_row_separator"
-        case navigationTitle = "navigation_title"
+        case listRowInsets
+        case listRowSeparator
+        case navigationTitle
```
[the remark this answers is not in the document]
carson-katri: Ok, however note that the Elixir lib converts them to camel case: https://github.com/liveviewnative/live_view_native_swift_ui/blob/3582c8ddbdd315914e5698f9ff6d393f4631253b/lib/live_view_native_swift_ui/modifiers.ex#L12
Perhaps that should be reverted as well.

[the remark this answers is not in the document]
shadowfacts: Ah, I missed that. Camel case is fine, then, so long as they're consistent with the backend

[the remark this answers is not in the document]
carson-katri: Personally I think the backend should leave them snake case, and the client can read with the proper keys.

Sources/LiveViewNative/LiveView.swift
carson-katri: I've made it opt-in via `NavigationMode.splitView`

carson-katri:
This is not quite ready yet, `NavigationLink` needs to be fixed.

Sources/LiveViewNative/Views/PhxHeroView.swift
```diff
@@ -1,166 +1,166 @@
...
 //
-//  PhxHeroView.swift
-//  LiveViewNative
-//
-//  Created by Shadowfacts on 5/2/22.
-//
-
-import SwiftUI
-
-struct PhxHeroView<R: CustomRegistry>: View {
```
carson-katri, line 10: @shadowfacts Is `PhxHeroView` something that should be in this package? I can't tell what SwiftUI functionality it's wrapping.

shadowfacts: No, it shouldn't be anymore. Originally it was part of the hero navigation transition, but that's been non-functional since iOS 16. If it returns, it'll either be as a wrapper for `.matchedGeometryEffect` (if that eventually supports nav transitions) or as a separate package if we're reimplementing it ourselves.

carson-katri: Ok, I'll remove it completely.

shadowfacts (approved):

PR #132, opened 2023-01-10
Snapshot Testing

carson-katri:
This adds a new method for testing an expected View tree against the built output.

Here's an example of a test:

https://github.com/liveviewnative/liveview-client-swiftui/blob/2cb9ddbb7b1ff919ad969ecbc4eb0e3a419f8d08/Tests/RenderingTests/TextTests.swift#L14-L18

You provide the markup and an expected View, and it generates snapshots and compares them. If the snapshots match the test will pass, otherwise it will save the images for the developer to compare.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -37,5 +37,9 @@ let package = Package(
         .testTarget(
             name: "LiveViewNativeTests",
             dependencies: ["LiveViewNative"]),
+        .testTarget(
+            name: "RenderingTests",
+            dependencies: ["LiveViewNative"]
+        )
     ]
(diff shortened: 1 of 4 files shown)


shadowfacts (approved):
We'll have to see how reliable this is long term—I've seen image comparison testing like this cause massive headaches, but this is definitely worth a try.

carson-katri:
I hope this is more resilient than some image comparison because it doesn't save any snapshots, they are generated from conceptually identical SwiftUI hierarchies if the test is successful.

PR #135, opened 2023-01-12
Add `TextField`/`SecureField`

carson-katri:
This re-implements `TextField` to use `SwiftUI.TextField` instead of wrapping UIKit. I also reorganized the View files a bit to match the various categories.

`SecureField` and `TextField` use a protocol to share functionality, but `SecureField` supports fewer options.

Here's an example using various field configuration options:

```html
<textfield
  placeholder="Send a message"
  name="body"
  text-field-style="rounded-border"
  keyboard="twitter"
  submit-label="send"
  phx-focus="body_focus"
  phx-blur="body_blur"
  autocapitalization="sentences"
  autocorrection="yes"
  axis="vertical"
/>
```


Sources/LiveViewNative/Views/Text Input and Output/TextFieldProtocol.swift
```diff
@@ -0,0 +1,216 @@
...
+            return .next
+        case "continue":
+            return .continue
+        default:
+            return nil
+        }
+    }
+}
+
+enum PhxTextFieldStyle: String {
```
shadowfacts, line 164: We're moving away from the Phx prefix, can you rename this and the various View extensions?

shadowfacts (changes requested):
Looks good, just one naming-related change

Sources/LiveViewNative/Views/Text Input and Output/TextFieldProtocol.swift
carson-katri: I removed the prefix from the enum, as well as the `PhxButtonStyle` enum. For the `View` extensions, I used the prefix `apply`, such as `applyTextFieldStyle` because of the name collision with the built-in modifier.

shadowfacts (approved):

PR #138, opened 2023-01-17
Cache `navigationTitle` for smoother transitions

carson-katri:
Stores the navigation title of a given `NavStackEntryView` so when navigating the back button retains the proper text:

## Before
Notice how the back button changes to the same value as the current title.

https://user-images.githubusercontent.com/13581484/212965776-7b27ba0c-ede4-49ad-ad99-7a3deef3e0de.mp4

## After
With these revisions, the back button is properly retained.

https://user-images.githubusercontent.com/13581484/212965787-873586e2-3ddb-4f50-b053-f0b2c3778696.mp4


carson-katri:
I simplified the `NavigationLink` implementation, because the other parameters weren't being used anymore. The markup would now be:

```html
<navigationlink href="/some/path">
  Click Me
</navigationlink>
```

shadowfacts (approved):

shadowfacts:
I agree, I think `destination` is clearer and better fits with SwiftUI. Calling it `value` definitely wouldn't be helpful.

PR #140, opened 2023-01-17
GitHub Actions workflow for running tests

carson-katri:
This adds a simple CI workflow that runs tests on an iOS simulator.


carson-katri:
@shadowfacts Looks like the build failed because there is no x86 build of the `liveview-native-core` framework? I imagine this may trip up users without M1/2 as well:
https://github.com/liveviewnative/liveview-client-swiftui/actions/runs/3942790937/jobs/6746857313#step:3:1086

shadowfacts:
I've pushed a new version of liveview-native-core that includes the x86_64 artifacts, lets see if that does the trick...

shadowfacts:
Well, the tests failed but they did indeed run, so things should be all set on the Intel front.

carson-katri:
I switched to a UIKit-based snapshot so we can test Views that aren't rendered by SwiftUI.

Tests are also passing now.

shadowfacts (approved):

PR #146, opened 2023-01-18
Add Form

### LINKED ISSUE #115: SwiftUI -> Layout Containers -> Forms: Form
Doc: https://developer.apple.com/documentation/swiftui/form

shadowfacts:
Closes #115

### DIFF
--- a/Sources/LiveViewNative/Views/Layout Containers/Forms/Form.swift
+++ b/Sources/LiveViewNative/Views/Layout Containers/Forms/Form.swift
@@ -1,0 +1,46 @@
+//
+//  Form.swift
+//  LiveViewNative
+//
+//  Created by Shadowfacts on 1/18/23.
+//
+
+import SwiftUI
+
+struct Form<R: CustomRegistry>: View {
+    @ObservedElement private var element
+    private let context: LiveContext<R>
+    
+    init(context: LiveContext<R>) {
+        self.context = context
+    }
+    
+    var body: some View {
+        SwiftUI.Form {
+            context.buildChildren(of: element)
+        }
+        .applyFormStyle(formStyle)
+    }
+    
+    private var formStyle: FormStyle {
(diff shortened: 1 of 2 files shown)


Sources/LiveViewNative/Views/Layout Containers/Forms/Form.swift
```diff
@@ -0,0 +1,53 @@
...
+            return .columns
+        case "grouped":
+            return .grouped
+        default:
+            return .automatic
+        }
+    }
+}
+
+private enum FormStyle {
```
carson-katri, line 37: `ButtonStyle` and `TextFieldStyle` use a `String` raw value instead of switching manually over the cases. Could be more resilient.

```suggestion
    private var formStyle: FormStyle {
        element.attributeValue(for: "form-style").flatMap(FormStyle.init) ?? .automatic
    }
}

private enum FormStyle: String {
```

Tests/RenderingTests/Layout Containers/Forms/FormTests.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+
+import XCTest
+import SwiftUI
+
+@MainActor
+final class FormTests: XCTestCase {
+
+    func testForm() throws {
+        try assertMatch("""
```
carson-katri, line 15: TextField cannot be rendered by SwiftUI's `ImageRenderer`, so this test will produce images like the following:

FormTests swift-15-testForm()_view

I can look into using a UIKit-based snapshot technique that can capture UIKit-wrapped Views in a separate PR.

shadowfacts: Ah, that's too bad. In that case, I'll remove this test for now until we revisit it.

carson-katri (approved):

Issue #142, opened 2023-01-17
[RFC] Syntax for Views with multiple child builders


carson-katri:
Some Views have a simple structure, such as `Button`:
```html
<button phx-click="some_action">
  <text>Child content goes here</text>
</button>
```

However, other Views such as [`Label<Title, Icon>`](https://developer.apple.com/documentation/swiftui/label) have multiple closures for children:

```swift
Label {
  // Title content goes here
} icon: {
  // Icon content goes here
}
```

This issue is to discuss the best syntax for this, and to track which Views depend on this functionality. Here's one possibility:

## Phoenix Components
Components can have [named slots](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#module-named-slots) with the following syntax:
```html
<.label>
  <:title>
    <text>Title content goes here</text>
  </:title>
  <:icon>
    <text>Icon content goes here</text>
  </:icon>
</.label>
```
Or with `title` being the default slot:
```html
<.label>
  <text>Title content goes here</text>
  <:icon>
    <text>Icon content goes here</text>
  </:icon>
</.label>
```

### Benefits
* Familiar syntax for Phoenix/LiveView users

### Drawbacks
* Components need to be declared on the Elixir side
  * It could be mostly automated with a macro to define which views are components and what slots they support
  * Alternatively: support slots for normal elements

shadowfacts:
This is a good option. Another possibility that doesn't require backend support is just using plain elements within the multi-child element, e.g.:
```html
<label>
  <title>
    <text>Hello</text>
  </title>
  <icon>
    <image system-name="globe" />
  </icon>
</label>
```

carson-katri:
Only thing to look out for there is name collisions if we also support a "default slot". For example, `ProgressView` has a `Label` child, which would be written like so:
```html
<progressview>
  <label> ... </label>
  <currentvaluelabel> ... </currentvaluelabel>
</progressview>
```
Which makes `<label>` ambiguous with the `Label` view, in the case where we support default slots.

carson-katri:
Namespacing is one option to avoid ambiguity without using slots.
```html
<menu>
  <menu:label>Pick one</menu:label>
  <menu:content>
    <button>Option 1</button>
    <button>Option 2</button>
  </menu:content>
</menu>
```
```swift
Menu {
  Button("Option 1") {}
  Button("Option 2") {}
} label: {
  Text("Pick one")
}
```

bcardarella:
[...]

PR #150, opened 2023-01-24
Add `Menu` and prototype for slot syntax

### LINKED ISSUE #67: SwiftUI -> User Interface Element -> Controls and indicator -> Menus: Menu
Doc: https://developer.apple.com/documentation/swiftui/menu

carson-katri:
Closes #67

This uses the prefixed slot syntax discussed in #142. If/when a custom engine is added, the following could be refactored to allow for a more familiar slot-like syntax (without any changes to the client code):

```html
<menu>
  <menu:label>Change Mode</menu:label>
  <menu:content>
    <%= for item <- 0..5 do %>
      <button>Option #<%= item %></button>
    <% end %>
  </menu:content>
</menu>
```

The engine would synthesize the above when a slot is used outside of a component by looking up at the nearest parent element.

```html
<menu>
  <:label>Change Mode</:label>
  <:content>
    <%= for item <- 0..5 do %>
      <button>Option #<%= item %></button>
    <% end %>
  </:content>
</menu>
```

https://user-images.githubusercontent.com/13581484/214320105-13ff1f22-16fe-414c-b02a-1d237be47a1a.mov

### DIFF
--- a/Sources/LiveViewNative/LiveContext.swift
+++ b/Sources/LiveViewNative/LiveContext.swift
@@ -64,14 +64,14 @@ public struct LiveContext<R: CustomRegistry> {
                 return false
             }
         })
-        let defaultSlotChildren = children.filter({
-            if case let .element(element) = $0.data {
-                return element.namespace != tagName
-            } else {
-                return true
-            }
-        })
         if namedSlotChildren.isEmpty && includeDefaultSlot {
+            let defaultSlotChildren = children.filter({
+                if case let .element(element) = $0.data {
+                    return element.namespace != tagName
+                } else {
+                    return true
+                }
+            })
             return ForEach(defaultSlotChildren.map({ ($0.id, $0.children()) }), id: \.0) { subChildren in
                 coordinator.builder.fromNodes(subChildren.1, context: self)
             }
--- a/Sources/LiveViewNative/LiveContext.swift
+++ b/Sources/LiveViewNative/LiveContext.swift
@@ -6,6 +6,7 @@
 //
 
 import SwiftUI
+import LiveViewNativeCore
 
 /// The context provides information at initialization-time to views in a LiveView.
 public struct LiveContext<R: CustomRegistry> {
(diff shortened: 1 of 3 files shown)


carson-katri:
Adding tests depends on #140 (for UIKit-based snapshots)

Sources/LiveViewNative/LiveContext.swift
```diff
@@ -45,4 +46,40 @@ public struct LiveContext<R: CustomRegistry> {
...
+            } else {
+                return false
+            }
+        })
+    }
+    
+    public func buildChildren(of element: ElementNode, defaultSlotFor tagName: String) -> some View {
+        buildChildren(of: element, where: {
+            if case let .element(element) = $0.data {
+                return element.namespace != "menu"
```
shadowfacts, line 79: This should presumably be `tagName`, not hardcoded

Sources/LiveViewNative/Views/Controls and Indicators/Menus/Menu.swift
```diff
@@ -0,0 +1,46 @@
...
+    init(element: ElementNode, context: LiveContext<R>) {
+        self.context = context
+    }
+    
+    public var body: some View {
+        SwiftUI.Menu {
+            context.buildChildren(of: element, withTagName: "content", namespace: "menu")
+        } label: {
+            context.buildChildren(of: element, defaultSlotFor: "menu")
+            context.buildChildren(of: element, withTagName: "label", namespace: "menu")
```
shadowfacts, line 23: I think these should be combined into one helper that uses the explicit version if present and the default slot otherwise. I can't think of a reason you'd want both the default and explicit one together

shadowfacts (changes requested):
I had a couple minor changes, but I think this is a solid approach for describing multi-children views in templates

shadowfacts (approved):

shadowfacts (approved):

PR #151, opened 2023-01-24
Fix macOS build

carson-katri:
Disables features unavailable on macOS.


bcardarella:
Can we expand this to a larger feature for all platforms?

carson-katri:
Sure, I'll look into doing that. It'll probably require rebuilding `liveview-native-core` for those platforms.

bcardarella:
@carson-katri is that because of Intel?

carson-katri:
It looks like it only builds for iOS/macOS in the Makefile: https://github.com/liveviewnative/liveview-native-core/blob/37d7970aaa3e0be32b40bf9ed895416ae17fc79b/Makefile.toml#L117

bcardarella:
Ah ok I'll check into this

carson-katri:
I added support for watchOS. It depends on https://github.com/liveviewnative/liveview-native-core/pull/5 and a new release of the xcframework (@shadowfacts is this automated?)

carson-katri:
Ok, thanks. I'll get that uploaded.

shadowfacts (approved):

carson-katri:
@shadowfacts The new version is released, so watchOS/macOS builds should be good to go

shadowfacts:
Awesome, this looks good to me :+1:

carson-katri:
I updated the workflow to make sure the package at least builds on macOS/watchOS. No tests run for them atm.

I also changed the saving of failed snapshots, they now are saved to the Report navigator in Xcode instead of writing to a temp directory.

PR #152, opened 2023-01-24
Add defaults to `CustomRegistry` associated types

carson-katri:
This adds default values to the associated types of `CustomRegistry` so the user does not have to manually specify `EmptyRegistry.None` values for the features they do not want to override.

You can now implement a registry that only changes a few features very easily:

```swift
/// Only changes the `loadingView`
struct MyAppRegistry: CustomRegistry {
    static func loadingView(for url: URL, state: LiveSessionCoordinator<Self>.State) -> some View {
        ProgressView(url.path(percentEncoded: false))
    }
}
```

Before, the same change would've required several typealiases to compile:

```swift
struct MyAppRegistry: CustomRegistry {
    typealias TagName = EmptyRegistry.None
    typealias ModifierType = EmptyRegistry.None
    typealias CustomView = Never
    
    static func loadingView(for url: URL, state: LiveSessionCoordinator<Self>.State) -> some View {
        ProgressView(url.path(percentEncoded: false))
    }
}
```


shadowfacts (approved):

shadowfacts:
@carson-katri Can you also update the examples in the docs to remove the now-unnecessary typealiases?

shadowfacts (approved):

PR #154, opened 2023-01-24
Consistent naming convention

carson-katri:
Uses the following naming scheme across all Views:

`MyViewName` -> `my-view-name`

Affected views include `v-stack`, `scroll-view`, `navigation-link`, and others.

### DIFF
--- a/Sources/LiveViewNative/BuiltinRegistry.swift
+++ b/Sources/LiveViewNative/BuiltinRegistry.swift
@@ -14,35 +14,35 @@ struct BuiltinRegistry {
     @ViewBuilder
     static func lookup<R: CustomRegistry>(_ name: String, _ element: ElementNode, context: LiveContext<R>) -> some View {
         switch name {
-        case "textfield":
+        case "text-field":
             TextField<R>(element: element, context: context)
-        case "securefield":
+        case "secure-field":
             SecureField<R>(element: element, context: context)
         case "text":
             Text(context: context)
-        case "hstack":
+        case "h-stack":
             HStack<R>(element: element, context: context)
-        case "vstack":
+        case "v-stack":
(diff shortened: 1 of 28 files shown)


shadowfacts (changes requested):
Can you update the docs as well? Should be a straightforward find & replace

Sources/LiveViewNative/BuiltinRegistry.swift
```diff
@@ -14,43 +14,43 @@ struct BuiltinRegistry {
...
-        case "textfield":
+        case "text-field":
             TextField<R>(element: element, context: context)
-        case "securefield":
+        case "secure-field":
             SecureField<R>(element: element, context: context)
         case "text":
             Text(context: context)
-        case "hstack":
+        case "h-stack":
```
bcardarella, line 23: I do wonder if we should alias this to `hstack` as well as the other `-stack` views. I'm all for consistency but in the case of these specific views it does feel better writing:

```html
<hstack>
...
</hstack>
```

rather than

```html
<h-stack>
...
</h-stack>
```

Thoughts? So the idea would be to preserver the `h-stack` but offer the nicer alternative

shadowfacts: For the stacks in particular, I'm fine permitting both options. `hstack` does read nicer than `h-stack`, and stacks are very common

bcardarella: yeah I don't think this should be allowed beyond stacks

carson-katri: I enabled this for stacks.

shadowfacts (approved):

PR #156, opened 2023-01-26
Disconnect live reload socket on disconnect

carson-katri:
This is a minor change to fix an issue with connections being left open when a live reload occurs. The connection list would continue growing every time the app received a reload. Now it stays around 3 connections.

There still seems to be an issue somewhere with leaving connections open, but I haven't found that yet and this seems to fix part of the problem.

### Before
Every time I cause a reload the list continues to grow.

https://user-images.githubusercontent.com/13581484/214864816-91c93465-c7b4-4bd6-a5c5-a61fc884f2f5.mov

### After
Slightly better, but some connections are still being left open.

https://user-images.githubusercontent.com/13581484/214865136-fd43479a-b497-4a7f-b0cd-b4aeaa80a4db.mov

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -135,6 +135,8 @@ public class LiveSessionCoordinator<R: CustomRegistry>: ObservableObject {
         navigationPath.removeAll()
         self.socket?.disconnect()
         self.socket = nil
+        self.liveReloadSocket?.disconnect()
+        self.liveReloadSocket = nil
         // when deliberately disconnect, don't let pending connections continue
         internalState = .notConnected(reconnectAutomatically: false)
     }


carson-katri:
Simplified things a bit, now it seems to always have at most 3 connections.

bcardarella:
@carson-katri I asked Chris McCord to check this out, I want to confirm that we're aligned with the upstream LV WS lifecycle expectations

carson-katri:
Ok, sounds good. So far with these changes I've had no problems with live reload even when just spamming save 😅

bcardarella (approved):

bcardarella:
@KronicDeth will your channels work in Core replace the need for us to fix this in Swift client?

carson-katri:
Ah ok, I'll revise it based on that

bcardarella:
@carson-katri I think the work that @KronicDeth for Channels in Core will replace all of this. Unless it is presenting a bug that is blocking you on other things it may be best to punt

shadowfacts:
I agree, once core is managing the LiveView channels, it should also inherit responsibility for live reload.

PR #161, opened 2023-01-31
Avoid `AnyView` in modifiers

carson-katri:
This refactors modifier handling to avoid the use of `AnyView`, which could impact performance.

Now, modifiers work more like elements. Custom registries should provide an `applyModifier(_:modifier:to:context:)` method to register custom modifier types:

```swift
struct MyRegistry: CustomRegistry {
  static func applyModifier(_ type: ModifierType, modifier: Modifier<Self>, to content: some View, context: LiveContext<Self>) -> some View {
    switch type {
    case .myModifier:
      content.modifier(MyCustomModifier(from: modifier))
    }
  }
}
```

Note that this will require some changes to the serialization in `live_view_native_swift_ui` (https://github.com/liveviewnative/live_view_native_swift_ui/pull/5), so that each parameter value is encoded as a string, similar to element attributes. These can then be decoded individually as needed.

Let me know your thoughts on these changes.

### DIFF
--- a/Sources/LiveViewNative/BuiltinRegistry.swift
+++ b/Sources/LiveViewNative/BuiltinRegistry.swift
@@ -7,6 +7,8 @@
 
 import SwiftUI
 
+/// This protocol provides access to the `ViewModifier` type returned from `decodeModifier` in the `BuiltinRegistry`.
+/// That type is used by `ModifierContainer.builtin`.
 protocol BuiltinRegistryProtocol {
     associatedtype BuiltinModifier: ViewModifier
     associatedtype ModifierType: RawRepresentable where ModifierType.RawValue == String
--- a/Sources/LiveViewNative/BuiltinRegistry.swift
+++ b/Sources/LiveViewNative/BuiltinRegistry.swift
@@ -7,7 +7,13 @@
 
 import SwiftUI
 
-struct BuiltinRegistry {
+protocol BuiltinRegistryProtocol {
+    associatedtype BuiltinModifier: ViewModifier
+    associatedtype ModifierType: RawRepresentable where ModifierType.RawValue == String
+    static func decodeModifier(_ type: ModifierType, from decoder: Decoder) throws -> BuiltinModifier
+}
+
+struct BuiltinRegistry: BuiltinRegistryProtocol {
     
     static let attributeDecoder = JSONDecoder()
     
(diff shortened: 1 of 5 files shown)


shadowfacts:
I'm fully in favor of removing the need for `AnyView` when applying modifiers,  but I'm not sure about this approach. I don't like that all the modifier arguments have to be encoded as strings. That was one of the big pitfalls of the original attributes-based approach, and I'd rather not go back.

carson-katri:
Ok, makes sense. I'll look into some other ways of accomplishing this.

shadowfacts:
The crux of the issue, I think, is that we need to decode the modifier to a concrete but unknown type without introducing an existential type so that the concrete type can still be applied later. I think we can work around this by essentially doing the same thing that SwiftUI does for `View`s:

Create a `@ViewModifierBuilder` result builder that can handle conditionals and wrap them in a single, concrete `ViewModifier` type, like `ViewBuilder` does for views. Something like the example below. Then, `CustomRegistry` could have another associated type for the modifier result that could be returned from `decodeModifier` and stored by the applicator view.

```swift
@resultBuilder
struct ViewModifierBuilder {
    static func buildEither<First: ViewModifier, Second: ViewModifier>(first component: First) -> ConditionalModifier<First, Second> {
        return .first(component)
    }
    // etc.
}
enum ConditionalModifier<First: ViewModifier, Second: ViewModifier>: ViewModifier {
    case first(First)
    case second(Second)
        switch self {
            // etc.
        }
    }
}
struct MyRegistry: CustomRegistry {
    func decodeModifier(type: ModifierType, from decoder: Decoder) -> some ViewModifier {
        switch type {
            case .thingOne:
                return ThingOne(from: decoder)
            // etc.
        }
    }
}
```

carson-katri:
I implemented this, and it seems to work well.

carson-katri:
Here's what a `CustomRegistry` looks like now:

```swift
struct MyRegistry: CustomRegistry {
    enum ModifierType: String {
        case background
        case spin
    }
    
    static func decodeModifier(_ type: ModifierType, from decoder: Decoder, context: LiveContext<Self>) throws -> some ViewModifier {
        switch type {
        case .background:
            try BackgroundModifier(from: decoder)
        case .spin:
            try SpinModifier(from: decoder)
        }
    }
    
    struct BackgroundModifier: ViewModifier, Decodable {
        let color: String
        
            content.background(Color(fromNamedOrCSSHex: color))
        }
    }
    struct SpinModifier: ViewModifier, Decodable {
        let speed: Double
        var clockwise: Bool = true
        
            TimelineView(.animation) { timeline in
                content.rotationEffect(.degrees(timeline.date.timeIntervalSince1970 * speed * (clockwise ? 1 : -1)))
            }
        }
    }
}
```

Sources/LiveViewNative/BuiltinRegistry.swift
```diff
@@ -7,7 +7,13 @@
 
 import SwiftUI
 
-struct BuiltinRegistry {
+protocol BuiltinRegistryProtocol {
```
shadowfacts, line 10: Am I right in my understanding that this protocol is necessary to tie the typealias to the opaque return type of the `decodeModifier` implementation, so that there's something that can be explicitly spelled out in the `ModifierContainer`?

If so, I'd add a brief comment to that effect

shadowfacts (approved):
Very nice work!

Sources/LiveViewNative/BuiltinRegistry.swift
carson-katri: Yep, exactly. I'll add a comment.

shadowfacts (approved):

PR #164, opened 2023-02-01
Live bindings and public event handling API

shadowfacts:
This PR adds two things:
1. A proper, public API for handling events on the client
2. The real meat and potatoes: live bindings, a mechanism for two-way client/server bindings

~~The only thing that remains before this can be merged is integrating live bindings with `@FormState`. Form state elements currently always use the `value` attribute and store data either locally or in an `<phx-form>`. Once this is implemented, if the `value-binding` attribute is present, it will establish a live binding which will be used for storage.~~

## Event Handling
Event handling uses Combine, which allows for cancelling previously created event handlers as well as easily listening for events from within the SwiftUI lifecycle (using `onReceive`).

## Live Bindings 🎉
Bindings must be declared on the server-side live view, in this example it's done like so:

```elixir
defmodule BackendtestWeb.TestLive do
  use BackendtestWeb, :live_view
  use BackendtestWeb.Bindings

  bindings(my_bool: false)

  def handle_event("toggle", _, socket) do
    {:noreply, set_my_bool(socket, !get_my_bool(socket))}
  end
end
```

And in the template it's used as:
```html
<text>Hello</text>
<text>my_bool: <%= @my_bool %></text>
<bound-toggle value="my_bool" />
<button phx-click="toggle">
	<text>Toggle from backend</text>
</button>
```

And the `bindings` macro, which will be part of the `live_view_native_swift_ui` package:
```elixir
defmacro bindings(names_and_defaults) do
  quote bind_quoted: [names_and_defaults: names_and_defaults] do
    on_mount {__MODULE__, :_set_binding_defaults}

    def on_mount(:_set_binding_defaults, _params, _session, socket) do
      {
        :cont,
        socket
        |> assign(unquote(names_and_defaults))
        |> push_event("_live_bindings", Map.new(unquote(names_and_defaults)))
      }
    end

    Enum.each(names_and_defaults, fn {name, _default} ->
      def unquote(:"get_#{name}")(socket) do
        socket.assigns[unquote(name)]
      end

      def unquote(:"set_#{name}")(socket, value) do
        socket
        |> assign(unquote(name), value)
        # todo: it would be nice if there were a hook that let us coalesce multiple changes in a single callback into one event
        |> push_event("_live_bindings", %{unquote(name) => value})
      end

      def handle_event(unquote(to_string(name)), %{"value" => value}, socket) do
        {:noreply, assign(socket, unquote(name), value)}
      end
    end)
  end
end
```

And finally from SwiftUI, the binding can be used like so:
```swift
struct BoundToggle: View {
    @LiveBinding(attribute: "value") private var value: Bool
    
    var body: some View {
        SwiftUI.Toggle(isOn: $value) {
            SwiftUI.Text("Bound toggle")
        }
    }
}
```

The screen recording shows both client -> server updates (when the toggle switch is clicked, and the text updates) and server -> client updates (when the button is clicked and the toggle switch updates) working:

https://user-images.githubusercontent.com/7091588/216186677-338e1f19-ad20-4f86-881f-586f7f17fa2b.mov

### From View Modifiers
A live binding can also be used on a view modifier. The property wrapper conforms to `Decodable`, so it can be initialized in the modifier's decode implementation.

### Form State
All `@FormState` elements now also have an optional `value-binding` attribute which specifies the name of a live binding to use in place of the local or `<phx-form>` storage.

### DIFF
--- a/Sources/LiveViewNative/NavStackEntryView.swift
+++ b/Sources/LiveViewNative/NavStackEntryView.swift
@@ -11,7 +11,7 @@ import LiveViewNativeCore
 struct NavStackEntryView<R: CustomRegistry>: View {
     private let entry: LiveNavigationEntry<R>
     @ObservedObject private var coordinator: LiveViewCoordinator<R>
-    @StateObject private var liveViewModel = LiveViewModel<R>()
+    @StateObject private var liveViewModel = LiveViewModel()
     
     init(_ entry: LiveNavigationEntry<R>) {
         self.entry = entry
@@ -35,6 +35,15 @@ struct NavStackEntryView<R: CustomRegistry>: View {
                 if let doc = newDocument {
                     // todo: doing this every time the DOM changes is probably not efficient
                     liveViewModel.updateForms(nodes: doc[doc.root()].depthFirstChildren())
+                }
+            }
+            .onReceive(coordinator.receiveEvent("_live_bindings"), perform: liveViewModel.updateBindings)
+            .onReceive(liveViewModel.bindingUpdatedByClient) { (name, value) in
+                // todo: consider grouping these into a single even per runloop iteration?
+                Task {
+                    try? await coordinator.pushEvent(type: "_live_bindings", event: name, value: [
+                        "value": value
+                    ])
                 }
             }
     }
--- a/Sources/LiveViewNative/NavStackEntryView.swift
+++ b/Sources/LiveViewNative/NavStackEntryView.swift
@@ -38,12 +38,12 @@ struct NavStackEntryView<R: CustomRegistry>: View {
                 }
             }
             .onReceive(coordinator.receiveEvent("_live_bindings"), perform: liveViewModel.updateBindings)
-            .onReceive(liveViewModel.bindingUpdatedByClient) { (name, value) in
-                // todo: consider grouping these into a single even per runloop iteration?
+            .onReceive(
+                liveViewModel.bindingUpdatedByClient
+                    .collect(.byTime(RunLoop.main, RunLoop.main.minimumTolerance))
+            ) { updates in
                 Task {
-                    try? await coordinator.pushEvent(type: "_live_bindings", event: name, value: [
-                        "value": value
-                    ])
+                    try? await coordinator.pushEvent(type: "_live_bindings", event: "_live_bindings", value: Dictionary(updates, uniquingKeysWith: { cur, new in new }))
                 }
             }
     }
--- a/Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
+++ b/Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
@@ -16,20 +16,23 @@ import Combine
 ///
 /// ### Binding Names
 /// The name of a live binding is not defined directly by the client.
-/// Instead, it always controlled by the backend, to prevent the name getting out-of-sync, especially when multiple client versions may be in use.
+/// Instead, it is always controlled by the backend, to prevent the name getting out-of-sync, especially when multiple client versions may be in use.
 ///
 /// Depending on how the live binding is used, the name is obtained in one of two ways:
 /// 1. If it is being used as part of an element, the binding name is provided by an attribute on the element.
 /// Pass the name of the attribute which specifies the binding name to the ``LiveBinding/init(attribute:)`` initializer.
 /// 2. If it is being used as part of a view modifier, the binding name is encoded as a string in the modifier payload.
-/// Decode the biding using the normal `Decodable` API, and the string value will be used as the binding name.
+/// Decode the binding using the ``init(decoding:in:initialValue:)`` initializer, and the string value at the given coding key will be used as the binding name.
 ///
 /// See below for examples of both use cases.
 ///
-/// ### Server Support
-/// As live bindings are two-way mechanism, they are not implemented purely on the client, but rather require server-side support.
+/// In either case, if the binding name is not provided, the `LiveBinding` operates in local mode (that is, like a regular SwiftUI `@State` property, not connected to the backend).
+/// When in local mode, an initial value for the property must be provided (see the respective initializers for how), otherwise accessing the live binding will crash.
 ///
-/// The server sends an event on mount with all of the binding values, sends an event when a binding is changed, and handles an even when a binding is changed on the client. These are all provided by the [`live_view_native_swift_ui`](https://github.com/liveviewnative/live_view_native_swift_ui) Elixir package.
+/// ### Server Support
+/// As live bindings are a two-way mechanism, they are not implemented purely on the client, but rather require server-side support.
+///
+/// The server sends an event on mount with all of the binding values, sends an event when a binding is changed, and handles an event when a binding is changed on the client. These are all provided by the [`live_view_native_swift_ui`](https://github.com/liveviewnative/live_view_native_swift_ui) Elixir package.
 ///
 /// The `bindings` macro takes a keyword list of bindings and their initial values and automatically generates all the necessary supporting code.
 /// The Elixir value used for the initial binding value must be serializable as JSON.
@@ -72,7 +75,7 @@ import Combine
 ///
 ///     init(from decoder: Decoder) throws {
 ///         let container = try decoder.container(keyedBy: CodingKeys.self)
-///         self._isActive = try container.decode(LiveBinding.self, forKey: .isActive)
+///         self._isActive = try LiveBinding(decoding: .isActive, in: container)
 ///     }
 ///
 ///     func body(content: Content) -> some View {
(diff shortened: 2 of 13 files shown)


carson-katri:
This looks great! Would live bindings replace `<phx-form>`?

shadowfacts:
Heh, I just edited my original comment. No, this won't replace `phx-form`. I think it still makes sense conceptually to have a group of form values which are used together for change/submit events.

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
```diff
@@ -0,0 +1,153 @@
...
+    private let attributeName: AttributeName
+    @StateObject private var data = Data()
+    
+    @ObservedElement private var element
+    @Environment(\.coordinatorEnvironment) private var coordinator
+    @EnvironmentObject private var liveViewModel: LiveViewModel
+    
+    private var bindingName: String {
+        guard let name = element.attributeValue(for: attributeName) else {
+            fatalError("@LiveBinding missing binding name for \(attributeName)")
```
carson-katri, line 70: If we were to use live bindings for things like `List.selection`, there would need to be a way to have the binding be optional. It's possible the user wouldn't use the selection. In that case, I think it should never change from the default value.

carson-katri:
Here's another idea for the Elixir-side, which I think would provide more flexibility.

```ex
defmodule BackendtestWeb.TestLive do
  use BackendtestWeb, :live_view
  use BackendtestWeb.Bindings

  # Each binding is it's own call with a type and default
  binding :my_bool, :boolean, default: false
  binding :counter, :integer, default: 5

  def handle_event("toggle", _, socket) do
    # Passing the name as an atom could give more flexibility, such as dynamically swapping out which is accessed.
    # That would be more complex if each binding has its own function.
    {:noreply, set_binding(socket, :my_bool, !get_binding(socket, :my_bool)}
    # Maybe store them under `assigns.bindings`?
    {:noreply, set_binding(socket, :my_bool, !socket.assigns.bindings.my_bool)}
  end
end
```

bcardarella:
@carson-katri we should pattern match to event handling, are you suggesting that `binding` expands out to this?

carson-katri:
`binding` wouldn't expand to that `handle_event`, that was an example of an event handler that updates the bindings.

bcardarella:
What does `binding` provide in this case? I might be missing some context here

carson-katri:
It serves the same purpose as the `bindings(...)` macro mentioned above, except it's split out so you can provide more information for each one such as type.

bcardarella:
@shadowfacts I think we should discuss this API. I'd prefer not to deviate from actual LV in how bindings are intended to be implemented in a Phoenix app

shadowfacts:
@bcardarella:
> I'd prefer not to deviate from actual LV in how bindings are intended to be implemented in a Phoenix app

These bindings are not the same as what regular Phoenix LiveView calls [bindings](https://hexdocs.pm/phoenix_live_view/bindings.html). Those are about binding backend handlers to client-side events. These are bindings in the [SwiftUI sense](https://developer.apple.com/documentation/swiftui/binding)—something that provides read and write access to a piece of data stored elsewhere. In this case, that data is shared between the client and the server which is what makes them live.

The "live bindings" name matches how SwiftUI thinks of these, but I'm open to calling it something else that doesn't conflict with existing LiveView notions.

@carson-katri:

> Each binding is it's own call with a type and default

That's definitely possible too, I just went with a single call because it made writing the macro (not my area of expertise) simpler. Having the types be specified would also let us better handle marshalling/unmarshalling the JSON that's sent in the event, rather than requiring the binding values themselves be JSON-compatible.

> Passing the name as an atom could give more flexibility, such as dynamically swapping out which is accessed.

Also makes sense.

> Maybe store them under `assigns.bindings`?

This I'm not sure about, as it makes accessing them in templates more cumbersome, and I'm not sure what's to be gained by having all the bindings together. A binding on the backend is fundamentally just a sparkling assign.

Sources/LiveViewNative/NavStackEntryView.swift
```diff
@@ -37,6 +37,15 @@ struct NavStackEntryView<R: CustomRegistry>: View {
                     liveViewModel.updateForms(nodes: doc[doc.root()].depthFirstChildren())
                 }
             }
+            .onReceive(coordinator.receiveEvent("_live_bindings"), perform: liveViewModel.updateBindings)
+            .onReceive(liveViewModel.bindingUpdatedByClient) { (name, value) in
+                // todo: consider grouping these into a single even per runloop iteration?
```
carson-katri, line 42: You might be able to accomplish that with
```swift
liveViewModel.bindingUpdatedByClient
  .collect(.byTime(RunLoop.main, RunLoop.main.minimumTolerance))
```

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
```diff
@@ -0,0 +1,215 @@
...
+///
+/// ### Binding Names
+/// The name of a live binding is not defined directly by the client.
+/// Instead, it always controlled by the backend, to prevent the name getting out-of-sync, especially when multiple client versions may be in use.
+///
+/// Depending on how the live binding is used, the name is obtained in one of two ways:
+/// 1. If it is being used as part of an element, the binding name is provided by an attribute on the element.
+/// Pass the name of the attribute which specifies the binding name to the ``LiveBinding/init(attribute:)`` initializer.
+/// 2. If it is being used as part of a view modifier, the binding name is encoded as a string in the modifier payload.
+/// Decode the biding using the normal `Decodable` API, and the string value will be used as the binding name.
```
carson-katri, line 25:
```suggestion
/// Decode the binding using the normal `Decodable` API, and the string value will be used as the binding name.
```

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
```diff
@@ -0,0 +1,215 @@
...
+import Combine
+
+/// Live bindings provide a mechanism for sharing state between the server and client in a way that can be updated by either.
+///
+/// Updates to the value from the client-side will automatically send an event to the server to update its stored value.
+/// Likewise, whenever the value is updated on the server, an event will be sent to the client which will trigger the `@LiveBinding` to update and the SwiftUI view that contains it (as well as any that use the ``projectedValue`` binding) to re-render.
+///
+/// ### Binding Names
+/// The name of a live binding is not defined directly by the client.
+/// Instead, it always controlled by the backend, to prevent the name getting out-of-sync, especially when multiple client versions may be in use.
```
carson-katri, line 19:
```suggestion
/// Instead, it is always controlled by the backend, to prevent the name getting out-of-sync, especially when multiple client versions may be in use.
```

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
```diff
@@ -0,0 +1,215 @@
...
+/// Depending on how the live binding is used, the name is obtained in one of two ways:
+/// 1. If it is being used as part of an element, the binding name is provided by an attribute on the element.
+/// Pass the name of the attribute which specifies the binding name to the ``LiveBinding/init(attribute:)`` initializer.
+/// 2. If it is being used as part of a view modifier, the binding name is encoded as a string in the modifier payload.
+/// Decode the biding using the normal `Decodable` API, and the string value will be used as the binding name.
+///
+/// See below for examples of both use cases.
+///
+/// ### Server Support
+/// As live bindings are two-way mechanism, they are not implemented purely on the client, but rather require server-side support.
```
carson-katri, line 30:
```suggestion
/// As live bindings are a two-way mechanism, they are not implemented purely on the client, but rather require server-side support.
```

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
```diff
@@ -0,0 +1,215 @@
...
+///
+/// The server sends an event on mount with all of the binding values, sends an event when a binding is changed, and handles an even when a binding is changed on the client. These are all provided by the [`live_view_native_swift_ui`](https://github.com/liveviewnative/live_view_native_swift_ui) Elixir package.
```
carson-katri, line 32:
```suggestion
/// The server sends an event on mount with all of the binding values, sends an event when a binding is changed, and handles an event when a binding is changed on the client. These are all provided by the [`live_view_native_swift_ui`](https://github.com/liveviewnative/live_view_native_swift_ui) Elixir package.
```

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
```diff
@@ -0,0 +1,215 @@
...
+    @EnvironmentObject private var liveViewModel: LiveViewModel
+    
+    private let bindingNameSource: BindingNameSource
+    private var bindingName: String {
+        switch bindingNameSource {
+        case .fixed(let name):
+            return name
+        case .attribute(let attr):
+            guard let name = element.attributeValue(for: attr) else {
+                fatalError("@LiveBinding missing binding name for \(attr)")
```
carson-katri, line 105: `LiveBinding` still cannot be optional. This could be changed fairly easily so that it can accept an optional default value which is then used if the attribute is not found. Then in the `nonmutating set` call `data.objectWillChange.send()` to ensure the value is not updating locally.

<details>
<summary>Here's the changes I made to `LiveBinding` to support this, so that it could work well with List's `selection` attribute.</summary>

```swift
@propertyWrapper
public struct LiveBinding<Value: Codable>: Decodable {
    @StateObject private var data = Data()
    
    @Environment(\.coordinatorEnvironment) private var coordinator
    @EnvironmentObject private var liveViewModel: LiveViewModel
    
    private let defaultValue: Value?
    
    private let bindingNameSource: BindingNameSource
    private var bindingName: String? {
        switch bindingNameSource {
        case .fixed(let name):
            return name
        case .attribute(let attr):
            return element.attributeValue(for: attr)
        }
    }
    
    var isBound: Bool {
        switch bindingNameSource {
        case .fixed(_):
            return true
        case .attribute(let attr):
            return element.attribute(named: attr) != nil
        }
    }
    
    /// Creates a `LiveBinding` property wrapper that uses the binding in the given attribute.
    ///
    /// This initializer should be used when the live binding is used as part of an element.
    /// See ``LiveBinding`` for a discussion of how the underlying binding name is determined and an example of this usage.
    public init(wrappedValue: Value? = nil, attribute: AttributeName) {
        self.bindingNameSource = .attribute(attribute)
        self.defaultValue = wrappedValue
    }
    
    /// Creates a `LiveBinding` by decoding its name from a container.
    ///
    /// This initializer should be used when the live binding is used as part of a view modifier.
    /// See ``LiveBinding`` for an example of how this is used.
    public init(from decoder: Decoder) throws {
        let container = try decoder.singleValueContainer()
        let name = try container.decode(String.self)
        self.bindingNameSource = .fixed(name)
        self.defaultValue = nil
    }
    
    /// The value of the binding.
    ///
    /// Setting this property will send an event to the server to update its value as well.
    public var wrappedValue: Value {
        get {
            guard let bindingName else { return defaultValue! }
            return data.getValue(from: liveViewModel, bindingName: bindingName)
        }
        nonmutating set {
            guard let bindingName else {
                data.objectWillChange.send()
                return
            }
            // update the local value
            data.mode = .local(newValue)
            // update the view model, which will send an update to the backend
            let encoder = FragmentEncoder()
            // todo: if encoding fails, what should happen?
            try! newValue.encode(to: encoder)
            liveViewModel.setBinding(bindingName, to: encoder.unwrap() as Any)
        }
    }
    
    /// A SwiftUI `Binding` that provides read and write access to the underlying data.
    ///
    /// Access this binding with the dollar sign prefix. If a property is declared as `@LiveBinding(name: "value") var value`, then the projected value binding can be access as `$value`.
    public var projectedValue: Binding<Value> {
        Binding {
            return self.wrappedValue
        } set: { newValue in
            self.wrappedValue = newValue
        }
    }
}
```

</details>

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
```diff
@@ -0,0 +1,215 @@
...
+        func getValue(from liveViewModel: LiveViewModel, bindingName: String) -> Value {
+            switch mode {
+            case .uninitialized:
+                cancellable = liveViewModel.bindingUpdatedByServer
+                    .filter { $0.0 == bindingName }
+                    .sink { [unowned self] _ in
+                        self.mode = .needsUpdateFromViewModel
+                        self.objectWillChange.send()
+                    }
+                fallthrough
```
carson-katri, line 195: It may be beneficial to refactor this to avoid `fallthrough`. I personally think it makes this harder to read and could lead to issues if someone inadvertently reorders the cases.

carson-katri (changes requested):
Just a few suggestions

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
shadowfacts: Ah, I see, you want to be able to use the LiveBinding even when a binding name isn't specified by the server. I was think of it as, in that case, you would have a separate State property. But I think you're right, it makes sense to fold that into LiveBinding.

shadowfacts: Or actually, is that not the case? Your code implies that LiveBinding is read-only and always returns the default value when a binding name isn't set.

I think a read/write local-only mode makes sense (I ran into needing that with TextEditor, and did this which isn't nice:
```swift
@State private var localIsFindPresented = false
@LiveBinding(attribute: "find-presented") private var liveIsFindPresented: Bool
var findPresented: Binding<Bool> { _liveIsFindPresented.isBound ? $liveIsFindPresented : $localIsFindPresented }
```

In that case, read/write access is necessary because passing `Binding.constant(false)` to the `findNavigator` modifier would, I believe, prevent it from being shown, which is not what we want if the binding name is omitted.

carson-katri (approved):
LGTM 👍

PR #165, opened 2023-02-02
Add Section

### LINKED ISSUE #111: SwiftUI -> Layout Containers -> Collection containers: Section
Doc: https://developer.apple.com/documentation/swiftui/section

carson-katri:
Closes #111 

```html
<section>
  <:header>5 Items</:header>
  <:content>
    <%= for item <- 1..5 do %>
      <text><%= item %></text>
    <% end %>
  </:content>
  <:footer>Above are some important items, in the range 1-5.</:footer>
</section>
```
The default slot can also be used for the content:
```html
<section>
  <:header>5 Items</:header>
  <%= for item <- 1..5 do %>
    <text><%= item %></text>
  <% end %>
  <:footer>Above are some important items, in the range 1-5.</:footer>
</section>
```

Simulator Screen Shot - iPhone 14 Pro - 2023-02-02 at 10 41 21

### Collapsible
On macOS, with the `sidebar` list style sections can be marked `collapsible`.

On iOS, sidebar list sections are always collapsible and the modifier is unavailable.

```html
<section collapsible="true">
  ...
</section>
<section> ... </section>
```

https://user-images.githubusercontent.com/13581484/216371451-70a03e8a-ea01-4e07-a762-9f98bf79fefc.mov

### DIFF
--- a/Sources/LiveViewNative/Views/Layout Containers/Collection Containers/Section.swift
+++ b/Sources/LiveViewNative/Views/Layout Containers/Collection Containers/Section.swift
@@ -18,10 +18,18 @@ struct Section<R: CustomRegistry>: View {
     public var body: some View {
         #if os(macOS)
         self.section
-            .collapsible(element.attributeValue(for: "collapsible").flatMap(Bool.init(_:)) ?? false)
+            .collapsible(isCollapsible)
         #else
         self.section
         #endif
+    }
+    
+    private var isCollapsible: Bool {
+        guard let collapsible = element.attribute(named: "collapsible")
+        else { return false }
+        guard let value = collapsible.value
+        else { return true }
+        return Bool(value) ?? false
     }
     
     private var section: SwiftUI.Section<some View, some View, some View> {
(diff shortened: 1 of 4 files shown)


Sources/LiveViewNative/Views/Layout Containers/Collection Containers/Section.swift
```diff
@@ -0,0 +1,44 @@
...
+        #endif
+    }
+    
+    private var isCollapsible: Bool {
+        guard let collapsible = element.attribute(named: "collapsible")
+        else { return false }
+        guard let value = collapsible.value
+        else { return true }
+        return Bool(value) ?? false
+    }
```
carson-katri, line 33: This allows the following:
```html
<section collapsible>
  ...
</section>
```

@shadowfacts Is there already something to get an attribute as a boolean, and if it has no value treat it as true? If not, I can extract this out because it seems like it would be useful for other boolean attributes.

shadowfacts: There is not currently, but I agree that would be useful to extract out.

carson-katri: Interestingly, if we treat these as [boolean attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes) the spec states:

> The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.

LV automatically omits the attribute if `false` is passed, and passes a `nil` value if `true` is passed:

shadowfacts (approved):

Issue #159, opened 2023-01-27
Bindings


AZholtkevych:
According to Brian, we have to open the discussion regarding Bindings How we can map those into SwiftUI

https://hexdocs.pm/phoenix_live_view/bindings.html

AZholtkevych:
FYI @carson-katri

carson-katri:
Here are some ideas for how each binding could map to SwiftUI.

> **Note** `N/A` does not mean it isn't possible to support, but instead that it doesn't map 1:1 with something from SwiftUI.

|  Binding  |  SwiftUI equivalent  |
| --------- | -------------------- |
| _Params_ |  |
| `value-*` | N/A |
| _Click Events_ |  |
| `phx-click` | [`onTapGesture(count:perform:)`](https://developer.apple.com/documentation/swiftui/view/ontapgesture(count:perform:)) |
| `phx-click-away` | N/A |
| _Form Events_ |  |
| `phx-change` | N/A |
| `phx-submit` | N/A |
| `phx-feedback-for` | N/A |
| `phx-disable-with` | N/A |
| `phx-trigger-action` | N/A |
| `phx-auto-recover` | N/A |
| _Focus Events_ |  |
| `phx-blur` | [`@FocusState`](https://developer.apple.com/documentation/swiftui/focusstate) |
| `phx-focus` | [`@FocusState`](https://developer.apple.com/documentation/swiftui/focusstate) |
| `phx-window-blur` | [`ScenePhase`](https://developer.apple.com/documentation/swiftui/scenephase) |
| `phx-window-focus` | [`ScenePhase`](https://developer.apple.com/documentation/swiftui/scenephase) |
| _Key Events_ |  |
| `phx-keydown` | [`keyboardShortcut(_:)`](https://developer.apple.com/documentation/swiftui/view/keyboardshortcut(_:)-8liec) |
| `phx-keyup` | N/A |
| `phx-window-keydown` | [`keyboardShortcut(_:)`](https://developer.apple.com/documentation/swiftui/view/keyboardshortcut(_:)-8liec) |
| `phx-window-keyup` | N/A |
| `phx-key` | N/A |
| _DOM Patching_ |  |
| `phx-mounted` | [`onAppear(action:)`](https://developer.apple.com/documentation/swiftui/view/onappear(perform:)) |
| `phx-update` | N/A |
[...]

PR #166, opened 2023-02-02
Add basic Phoenix bindings support

carson-katri:
This is the beginning of support for the bindings referenced in #159 

I have added support for `phx-window-focus` and `phx-window-blur` to test the concept.

> **Note** Those particular events seem to be broken in LiveView currently: https://github.com/phoenixframework/phoenix_live_view/issues/2161

I don't believe this overlaps with #164.


Sources/LiveViewNative/BuiltinRegistry.swift
```diff
@@ -122,4 +123,58 @@ struct BuiltinRegistry: BuiltinRegistryProtocol {
...
+            ScenePhaseObserver(content: view) { scenePhase in
+                if scenePhase == .background {
+                    Task {
+                        try await context.coordinator.pushEvent(type: "blur", event: event, value: value)
+                    }
+                }
+            }
+        case "phx-click":
+            // Special case for `Button`, which handles this event itself.
+            if element.tag == "button" {
```
carson-katri, line 155: I'm not sure if there should be a more general purpose way of representing that an element handles the event internally. Is this rare enough to just handle the few edge cases?

Sources/LiveViewNative/BuiltinRegistry.swift
```diff
@@ -122,4 +123,90 @@ struct BuiltinRegistry: BuiltinRegistryProtocol {
...
+        case "phx-focus":
+            FocusObserver(content: view) { isFocused in
+                guard isFocused else { return }
+                Task {
+                    try await context.coordinator.pushEvent(type: "focus", event: event, value: value)
+                }
+            }
+        case "phx-blur":
+            // Special case for `TextFieldProtocol`, which handles this event itself.
+            if element.tag == "text-field" || element.tag == "secure-field" {
```
carson-katri, line 162: Another special case is needed for `phx-focus`/`phx-blur`, so this may be fairly common.

carson-katri:
I added a new property wrapper, `@Event`. It provides a mechanism for sending events with automatic debounce/throttle handling.
It requires the name of the attribute that holds the event name, and an event type. It will automatically check for `phx-debounce`/`phx-throttle` attributes.

```swift
struct MyCounterElement: View {
  @Event("on-increment", type: "click") private var increment

  var body: some View {
    MyCounterView(onIncrement: {
      increment(["new-value": 10]) {
        // on complete
      }
    })
  }
}
```
There's probably also a way to have the debounce/throttle checking happen on the `LiveViewCoordinator` side, but I'm not sure exactly how that would look.

@shadowfacts Let me know your thoughts on this.

Sources/LiveViewNative/Property Wrappers/ObservedElement.swift
```diff
@@ -54,6 +54,10 @@ public struct ObservedElement {
         }
         return element
     }
+    
+    public var projectedValue: AnyPublisher<Void, Never> {
```
shadowfacts, line 58: This can be `some Publisher<Void, Never>` to avoid the erasure

Sources/LiveViewNative/Property Wrappers/Event.swift
```diff
@@ -0,0 +1,115 @@
...
+        self.event = nil
+        self.name = name
+        self.type = type
+    }
+    
+    public init(_ name: String, type: String) {
+        self.event = nil
+        self.name = .init(name: name)
+        self.type = type
+    }
```
shadowfacts, line 93: I also don't think this initializer is necessary, again because of the ExpressibleByStringLiteral conformance

Sources/LiveViewNative/Property Wrappers/Event.swift
```diff
@@ -0,0 +1,115 @@
...
+    
+    public init(event: String, type: String) {
```
shadowfacts, line 95: I'd document how this initializer differs from the attribute name one

shadowfacts (changes requested):
I left some minor suggestions, but overall I like the `@Event` approach for accessing event handlers.

The one part of this PR I'm not a fan of is the special casing for views that handle certain events internally, like you mentioned. Perhaps the views that do so could communicate that up to the event handler wrappers with view preferences or something similar?

Though come to think of it, we don't want that information to propagate up out of the particular element. I'm not sure SwiftUI has a good mechanism for that :/

carson-katri:
Improved the event handling to send the `phx-target` value for the `"cid"` key. Now events in a live component will be handled correctly.

shadowfacts (approved):
Nice work!

PR #169, opened 2023-02-07
Add `ShareLink`

carson-katri:
This adds support for sharing content via the native share sheet.

Here is the syntax:
```html
<share-link
    item="https://dockyard.com"
    subject="Check out DockYard's website"
    message="Here's a link to the DockYard homepage"
>
    <share-link:preview title="DockYard Homepage">
        <preview:image><image name="dockyard" /></preview:image>
        <preview:icon><image name="dockyard" /></preview:icon>
    </share-link:preview>
</share-link>
```

Should `<share-link:preview>` be renamed to `<share-preview>`? It is both a type and an argument. It also makes `<preview:image>` more confusing because the parent is not clearly named `<preview>`. An alternative design would be:

```html
<share-link
    item="https://dockyard.com"
    subject="Check out DockYard's website"
    message="Here's a link to the DockYard homepage"
>
    <share-preview title="DockYard Homepage">
        <share-preview:image><image name="dockyard" /></preview:image>
        <share-preview:icon><image name="dockyard" /></preview:icon>
    </share-preview>
</share-link>
```

I chose to implement the preview as an element, not a custom type or attribute on `<share-link>` because it takes `Image` views as arguments, and we have no way of passing views into types on the Elixir side. 

https://user-images.githubusercontent.com/13581484/217379028-f6149f96-d973-406e-b0b8-81f498592411.mp4

## Sharing Multiple Items
Multiple items can be shared by passing an `items` parameter with a JSON-encoded list of items. You can also provide a separate preview for each item.
> The preview with the least options passed will force others to use fewer as well. So if one preview provides `title` and `image`, and another preview provides `title`, `image`, and `icon`, both will only provide the `title` and `image` options. This is due to a limitation in SwiftUI.
```html
<share-link
    items={@items |> Jason.encode!}
    subject="Check out DockYard's website"
    message="Here's a link to the DockYard homepage"
>
    <share-link:preview item="https://dockyard.com" title="DockYard">
        <preview:image><image name="dockyard" /></preview:image>
        <preview:icon><image name="dockyard" /></preview:icon>
    </share-link:preview>
    <share-link:preview item="https://news.ycombinator.com" title="Hacker News">
        <preview:image><image name="hackernews" /></preview:image>
        <preview:icon><image name="hackernews" /></preview:icon>
    </share-link:preview>
    <share-link:preview item="https://apple.com" title="Apple">
        <preview:image><image name="apple" /></preview:image>
        <preview:icon><image name="apple" /></preview:icon>
    </share-link:preview>
</share-link>
```

## TODOs
- [x] Support multiple items
- [ ] ~~Only text is supported. A URL is automatically inferred by the system if applicable. Should other `Transferable` types be supported? Maybe only the built-in/useful ones?~~


carson-katri:
According to the [CoreTransferable docs](https://developer.apple.com/documentation/coretransferable?changes=_3), the system types that support `Transferable` are [String](https://developer.apple.com/documentation/Swift/String?changes=_3), [Data](https://developer.apple.com/documentation/foundation/data?changes=_3), [URL](https://developer.apple.com/documentation/foundation/url?changes=_3), and [Image](https://developer.apple.com/documentation/SwiftUI/Image?changes=_3). I'm not sure how we should handle `Image`, but the others should be doable?

carson-katri:
The items need to be manually encoded to a JSON array. Could this have a special case in the engine once we have more flexibility there? Similar to how the LiveView `HTMLEngine` performs some transformations for the `style` and `class` attributes, we could automatically JSON encode this `items` attribute.

Sources/LiveViewNative/Views/Controls and Indicators/Links/ShareLink.swift
```diff
@@ -0,0 +1,328 @@
...
+    
+    /// The set of values used to create the `SharePreview` for each item.
+    private enum PreviewBranch<Item: Hashable> {
+        struct Properties<Property> {
+            /// A map between share items and a certain property of the `SharePreview`.
+            let values: [Item:Property]
+            /// The value in the default `SharePreview`.
+            let defaultValue: Property?
+            
+            @inlinable
```
shadowfacts, line 221: I don't think marking these inlinable does anything, since they're module-internal anyway. If you want them to be inlined, `@inline(__always)` is what you're after.

Sources/LiveViewNative/Views/Controls and Indicators/Links/ShareLink.swift
```diff
@@ -0,0 +1,328 @@
...
+            )
+        case [\.title, \.image, \.icon]:
+            return .complete(
+                .init(previews.0.mapValues(\.title), default: previews.default?.title),
+                images: .init(previews.0.compactMapValues(\.image), default: previews.default?.image),
+                icons: .init(previews.0.compactMapValues(\.icon), default: previews.default?.icon)
+            )
+        default:
+            return .titled(.init(previews.0.mapValues(\.title), default: previews.default?.title))
+        }
```
shadowfacts, line 326: This feels more complex than necessary? Could this just be something like:
```swift
let image = previews.0.values.allSatisfy({ $0.image != nil })
let icon = previews.0.values.allSatisfy({ $0.icon != nil })

if image && icon {
	// ...
} else if image {
	// ...
} else if icon {
	// ...
} else {
	// just title
}
```

shadowfacts (changes requested):
Re the question in your original comment: How about `<share-link-preview>` and then `<share-link-preview:icon>`, etc.? I think that makes it clear that the preview element is part of the share link, and not a regular element, and also tying the inner elements to the preview element.

Sources/LiveViewNative/Views/Controls and Indicators/Links/ShareLink.swift
carson-katri: Oops, thanks. Best to leave it to the compiler.

shadowfacts (approved):

Issue #96, opened 2023-01-03
SwiftUI -> Layout Containers -> Lazy stacks: ViewLayout: LazyVStack


AZholtkevych:
Doc: https://developer.apple.com/documentation/swiftui/lazyvstack

PR #175, opened 2023-02-09
Add Lazy Stacks

### LINKED ISSUE #95: SwiftUI -> Layout Containers -> Lazy stacks: LazyHStack
Doc: https://developer.apple.com/documentation/swiftui/lazyhstack

carson-katri:
Closes #95, #96 

Lazy stacks don't create items until they will be on screen. This can be seen in the demo below, where the `AsyncImage` doesn't load until the item is scrolled into view.

The `pinned-views` option automatically pins the `<section>` headers to the top of the scroll view, as can be seen in the demo as well.

https://user-images.githubusercontent.com/13581484/217910317-347c6c5b-93ed-4a44-9b34-7414e09447f0.mov

```html
<scroll-view>
  <lazy-vstack pinned-views="section-headers">
    <%= for {season, episodes} <- @episodes do %>
      <section id={season |> Integer.to_string}>
        <section:header>
          <vstack spacing="0">
            <divider />
            <zstack alignment="leading">
              <rectangle fill-color="system-white" modifiers={@native |> frame(height: 40)} />
                <hstack modifiers={@native |> padding(all: 8)}>
                  <text font="subheadline" color="system-secondary">Season <%= season %></text>
                </hstack>
            </zstack>
          </vstack>
        </section:header>
        <%= for episode <- episodes do %>
          <hstack id={episode["id"] |> Integer.to_string} modifiers={@native |> padding(all: 8)}>
            <async-image src={episode["image"]["medium"]} modifiers={@native |> frame(width: 160, height: 90)} />
              <vstack alignment="leading">
                <text font="title2" font-weight="semibold"><%= episode["name"] %></text>
                <text font="subheadline" color="system-secondary">S <%= episode["season"] %> E <%= episode["number"] %></text>
                <spacer />
              </vstack>
              <spacer />
              <link destination={episode["url"]}>
                <image system-name="info.circle" symbol-scale="large" />
              </link>
          </hstack>
        <% end %>
      </section>
    <% end %>
  </lazy-vstack>
</scroll-view>
```

### DIFF
--- a/Sources/LiveViewNative/BuiltinRegistry.swift
+++ b/Sources/LiveViewNative/BuiltinRegistry.swift
@@ -94,6 +94,8 @@ struct BuiltinRegistry: BuiltinRegistryProtocol {
         case "text-editor":
             TextEditor(element: element, context: context)
 #endif
+        case "lazy-v-stack", "lazy-vstack":
+            LazyVStack(element: element, context: context)
             
         case "phx-form":
             PhxForm<R>(element: element, context: context)
--- a/Sources/LiveViewNative/BuiltinRegistry.swift
+++ b/Sources/LiveViewNative/BuiltinRegistry.swift
@@ -96,6 +96,8 @@ struct BuiltinRegistry: BuiltinRegistryProtocol {
 #endif
         case "lazy-v-stack", "lazy-vstack":
             LazyVStack(element: element, context: context)
+        case "lazy-h-stack", "lazy-hstack":
+            LazyHStack(element: element, context: context)
             
         case "phx-form":
             PhxForm<R>(element: element, context: context)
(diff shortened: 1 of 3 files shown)


shadowfacts (commented):
The code looks good, but based on your screen recording it seems like view state is being lost when a view scrolls offscreen? I don't think—correct me if I'm wrong—that an AsyncImage in plain SwiftUI would have to reload if it was onscreen, then scrolled away, then came back onscreen.

carson-katri:
I just tested it and the behavior matches SwiftUI. If the `AsyncImage` is the only content it isn't invalidated, but if it's nested with another View it reloads when scrolled away and back. That behavior matched SwiftUI as well.

shadowfacts:
Huh, well in that case, it's not our problem.

shadowfacts (approved):

PR #180, opened 2023-02-14
List selection support

carson-katri:
This adds support for `List.selection` using `LiveBinding`. You can specify a String value or a set of String values as the selection.

```html
<list selection="selection">
    <%= for item <- 1..10 do %>
        <text id={item |> Integer.to_string}><%= item %></text>
    <% end %>
</list>
```

```ex
defmodule ... do
  bindings(selection: ["1", "3", "5"], single_value_selection: "1", empty_single_selection: nil)

  ...
end
```

https://user-images.githubusercontent.com/13581484/218798033-150c6b1e-b9c9-4baa-9636-3a899be3153e.mp4

If you start with a value of `nil` or a single string, it will only allow one row to be selected at a time.

### DIFF
--- a/Sources/LiveViewNative/Views/Layout Containers/Collection Containers/List.swift
+++ b/Sources/LiveViewNative/Views/Layout Containers/Collection Containers/List.swift
@@ -10,6 +10,9 @@ import SwiftUI
 struct List<R: CustomRegistry>: View {
     @ObservedElement private var element: ElementNode
     private let context: LiveContext<R>
+    #if !os(watchOS)
+    @Environment(\.editMode) var editMode
+    #endif
     
     /// Selection of either a single value or set of values.
     enum Selection: Codable {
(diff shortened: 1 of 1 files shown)


carson-katri:
I've included support for `onMove` with a `phx-move` event.

```html
<list selection="selection" phx-move="on_move">
    <%= for item <- @items do %>
        <text id={item |> Integer.to_string}><%= item %></text>
    <% end %>
</list>
```
```ex
def handle_event("on_move", %{ "index" => index, "destination" => destination }, socket) do
  # There may be a more efficient/idiomatic way of performing a move in Elixir that I don't know about.
  {element, list} = List.pop_at(socket.assigns.items, index)
  moved = List.insert_at(list, (if destination > index, do: destination - 1, else: destination), element)
  {:noreply, assign(socket, :items, moved)}
end
```

https://user-images.githubusercontent.com/13581484/218810733-d0369bdf-eadd-4d1b-a138-144f21782e17.mp4

shadowfacts (approved):

bcardarella:
I cannot immediately find the video but @chrismccord has a demo of LiveView drag/drop. I'll check with him if he is using `phx-move` or a different event name. I think we should try to align when possible.

PR #185, opened 2023-02-15
Don't use State property to store FormState local value

shadowfacts:
Fixes a crash when focusing a local-value text field.

Previously, focusing the text field would result in two things:
1. the binding being set (which would change the mode to .local and update the localValue property)
2. the FocusState changing which would trigger a view update

But when the FormState value was read as a result of #2, the mode would be .local, the localValue property would still have the old nil value and so the force-unwrap would crash.

I'm not exactly sure why localValue wouldn't return the new value in the view update, but regardless, storing the current value and the mode together keeps them from getting out of sync.


carson-katri (approved):
LGTM 👍

PR #186, opened 2023-02-15
Add Picker

### LINKED ISSUE #72: SwiftUI -> User Interface Element -> Controls and indicator -> Pickers: Picker
Doc: https://developer.apple.com/documentation/swiftui/picker

shadowfacts:
Closes #72 

Right now, the tag modifier (and thus picker values) are limited to strings. I'd like to allow more complex types or any FormValue-conforming type in the future, but SwiftUI is really picky about the type of the picker's selection binding and the value passed into `.tag` lining up (e.g., if the binding is for a `String?` and the type passed into tag is a non-optional `String`, SwiftUI will think that there are picker options without tags).

Also, the [multi-source Picker initializers](https://developer.apple.com/documentation/swiftui/picker/init(sources:selection:content:label:)) aren't supported, since LVN doesn't have a concept of passing unspecified types and keypaths over the wire.

```html
<%= for style <- ["inline", "menu", "navigation-link", "segmented"] do %>
  <picker value-binding="picker_value" picker-style={style}>
    <picker:label>
      <text>Pick an icon</text>
    </picker:label>
    <picker:content>
      <%= for name <- ["lasso.and.sparkles", "paperplane", "tray.full", "doc.richtext", "graduationcap", "ellipsis.bubble"] do %>
        <label system-image={name} modifiers={[%{type: "tag", value: name}] |> Jason.encode!()}><text><%= name %></text></label>
      <% end %>
    </picker:content>
  </picker>
<% end %>
```

https://user-images.githubusercontent.com/7091588/219177928-6f609777-a493-4690-8599-b4a64a42d761.mp4

### DIFF
--- a/Sources/LiveViewNative/BuiltinRegistry.swift
+++ b/Sources/LiveViewNative/BuiltinRegistry.swift
@@ -112,7 +112,8 @@ struct BuiltinRegistry: BuiltinRegistryProtocol {
             LazyVStack(element: element, context: context)
         case "lazy-h-stack", "lazy-hstack":
             LazyHStack(element: element, context: context)
-            
+        case "picker":
+            Picker(context: context)
         case "phx-form":
             PhxForm<R>(element: element, context: context)
         case "phx-submit-button":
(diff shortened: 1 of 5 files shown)


carson-katri (commented):
Looks good. Could you add a test to `PickerTests.swift`? Also, will you be implementing the tag modifier in `live_view_client_swift_ui` with the current modifier schema setup, or is that still in flux?

shadowfacts:
I've added the tests. I will add the modifier to the Elixir lib, but I wanted to wait until the repos were merged just to get that out of the way first.

carson-katri (approved):

PR #198, opened 2023-02-17
Assorted documentation updates

shadowfacts:
And a couple pieces of code-cleanup

### DIFF
--- a/Sources/LiveViewNative/LiveContext.swift
+++ b/Sources/LiveViewNative/LiveContext.swift
@@ -63,6 +63,7 @@ public struct LiveContext<R: CustomRegistry> {
         }
     }
     
+    /// Checks whether the element has any children with the given tag.
     public func hasChild(
         of element: ElementNode,
         withTagName tagName: String,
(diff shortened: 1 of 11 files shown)


Sources/LiveViewNative/LiveContext.swift
```diff
@@ -63,6 +63,7 @@ public struct LiveContext<R: CustomRegistry> {
         }
     }
     
+    /// Checks whether the element has any children with the give tag.
```
carson-katri, line 66:
```suggestion
    /// Checks whether the element has any children with the given tag.
```

Sources/LiveViewNative/LiveViewNative.docc/FormControls.md
```diff
@@ -50,4 +50,10 @@ struct MicToggle: View {
 }
 ```
 
-Use this view in your LiveView view tree using the ``CustomRegistry`` (see <doc:AddCustomElement>) inside of a `<phx-form>` element. It will participate in the form model automatically, and its value will be sent whenever change or submit events are triggered on the form.
+Use this view in your LiveView view tree using the ``CustomRegistry`` (see <doc:AddCustomElement>).
+
+If used inside of a `<phx-form>`, it will participate in the form model automatically and its value will be sent whenever change or submit events are triggered on the form. The initial value will be read from the `value` attribute on the element, if present.
+
+When used outside of `<phx-form>`, the value is stored directly by the `FormState`. The initial value will be read from the `value` attribute on the element, if present. The `@FormState` property wrapper will automatically detect changes to its value and send change events to the backend, if an event nmae has been provided in the `phx-change` attribute on the element.
```
carson-katri, line 57:
```suggestion
When used outside of `<phx-form>`, the value is stored directly by the `FormState`. The initial value will be read from the `value` attribute on the element, if present. The `@FormState` property wrapper will automatically detect changes to its value and send change events to the backend, if an event name has been provided in the `phx-change` attribute on the element.
```

Sources/LiveViewNative/LiveViewNative.docc/GettingStarted.md
```diff
@@ -17,29 +17,29 @@ The LiveView is then created by passing in the coordinator, no other setup neces
 ```swift
 @MainActor
 struct ContentView: View {
-    @State private var coordinator = LiveViewCoordinator(URL(string: "http://localhost:4000/")!)
+    @State private var session = LiveSessionConfiguration(URL(string: "http://localhost:4000/")!)
```
carson-katri, line 20:
```suggestion
    @State private var session = LiveSessionCoordinator(URL(string: "http://localhost:4000/")!)
```

carson-katri (changes requested):

carson-katri (approved):

PR #203, opened 2023-02-18
Add foreground style modifier, gradients and improved colors

supernintendo:
Adds support for the [`foregroundStyle`](https://developer.apple.com/documentation/swiftui/view/foregroundstyle(_:)) modifier and [`LinearGradient`](https://developer.apple.com/documentation/swiftui/lineargradient):

```heex
<vstack>
  <label modifiers={foreground_style(@native, primary: {:color, :mint})}>Color Text</label>
  <label modifiers={foreground_style(@native, primary: {:linear_gradient, %{gradient: {:colors, [:red, :orange, :yellow, :green, :blue, :indigo, :purple]}, start_point: {0, 0}, end_point: {1, 1}}})}>
    Foreground Linear Gradient
  </label>
  <label modifiers={foreground_style(@native, primary: {:linear_gradient, %{gradient: {:stops, [{:pink, 0.8}, {:blue, 0.9}]}}})}>
    Foreground Linear Gradient with Stops
  </label>
</vstack>
```

Screenshot 2023-03-03 at 1 35 36 PM

Also introduces a couple of non-breaking changes to [`Color`](https://developer.apple.com/documentation/swiftui/color) handling for all modifiers:
- Standard and semantic colors can now be written as atoms (i.e. `:blue`, `:teal`, `:primary` instead of `"system-blue", "system-teal", "system-primary"`).
- Creating colors from an [RGB color space](https://developer.apple.com/documentation/swiftui/color/rgbcolorspace) is now supported.

```heex
<vstack>
  <button id="tint_hex" modifiers={tint(@native, color: "#3EB489")}>Hex Code</button>
</vstack>
<vstack>
  <button modifiers={tint(@native, color: :black)}>Black</button>
  <button modifiers={tint(@native, color: :blue)}>Blue</button>
  <button modifiers={tint(@native, color: :brown)}>Brown</button>
  <button modifiers={tint(@native, color: :clear)}>Clear</button>
  <button modifiers={tint(@native, color: :cyan)}>Cyan</button>
  <button modifiers={tint(@native, color: :gray)}>Gray</button>
  <button modifiers={tint(@native, color: :green)}>Green</button>
  <button modifiers={tint(@native, color: :indigo)}>Indigo</button>
  <button modifiers={tint(@native, color: :mint)}>Mint</button>
  <button modifiers={tint(@native, color: :orange)}>Orange</button>
</vstack>
<vstack>
  <button modifiers={tint(@native, color: :pink)}>Pink</button>
  <button modifiers={tint(@native, color: :purple)}>Purple</button>
  <button modifiers={tint(@native, color: :red)}>Red</button>
  <button modifiers={tint(@native, color: :teal)}>Teal</button>
  <button modifiers={tint(@native, color: :white)}>White</button>
  <button modifiers={tint(@native, color: :yellow)}>Yellow</button>
  <button modifiers={tint(@native, color: :accent)}>Accent</button>
  <button modifiers={tint(@native, color: :primary)}>Primary</button>
  <button modifiers={tint(@native, color: :secondary)}>Secondary</button>
</vstack>
<vstack>
  <button modifiers={tint(@native, color: {:srgb, %{white: 0.5, opacity: 0.5}})}>sRGB Grayscale</button>
  <button modifiers={tint(@native, color: {:srgb, %{red: 0.25, green: 0.25, blue: 0.75, opacity: 0.5}})}>sRGB Red Green Blue</button>
  <button modifiers={tint(@native, color: {:srgb_linear, %{white: 0.5, opacity: 0.5}})}>sRGB (Linear) Grayscale</button>
  <button modifiers={tint(@native, color: {:srgb_linear, %{red: 0.25, green: 0.25, blue: 0.75, opacity: 0.5}})}>sRGB (Linear) Red Green Blue</button>
  <button modifiers={tint(@native, color: {:display_p3, %{white: 0.5, opacity: 0.5}})}>DCI-P3 Grayscale</button>
  <button modifiers={tint(@native, color: {:display_p3, %{red: 0.25, green: 0.25, blue: 0.75, opacity: 0.5}})}>DCI-P3 Grayscale</button>
</vstack>
```

Screenshot 2023-03-03 at 1 36 13 PM


Sources/LiveViewNative/Utils/WrappedShapeStyle.swift
```diff
@@ -0,0 +1,46 @@
+//
+//  WrappedShapeStyle.swift
+// LiveViewNative
+//
+//  Created by May Matyi on 2/17/23.
+//
+
+import SwiftUI
+
+struct WrappedShapeStyle: Decodable {
```
carson-katri, line 10: This may be better represented as an `enum`:

```swift
enum WrappedShapeStyle: Decodable {
  case color(Color)
  case linearGradient(LinearGradient)

  ...
}
```

It could even be an extension on [`AnyShapeStyle`](https://developer.apple.com/documentation/swiftui/anyshapestyle):

```swift
extension AnyShapeStyle: Decodable {
  init(from decoder: Decoder) throws {
    ...
    // if it's a color
    self = Self(color)
    // if it's a gradient
    self = Self(gradient)
  }
}
```
If you go with an extension, there would be no need to switch over the `concreteStyle` in the modifiers, you can just use the decoded style directly.

lib/live_view_native_swift_ui/types/color.ex
```diff
@@ -1,28 +1,64 @@
...
-  def cast({:ns_color, ns_color}), do: {:ok, %__MODULE__{type: :ns_color, ns_color: ns_color}}
-  def cast({:cg_color, cg_color}), do: {:ok, %__MODULE__{type: :cg_color, cg_color: cg_color}}
+
+  def cast(value) when is_bitstring(value),
+    do: {:ok, %__MODULE__{create_with: :string, string: value}}
+
+  def cast(value) when value in @system_colors,
+    do: {:ok, %__MODULE__{create_with: :string, string: "system-#{value}"}}
+
+  def cast({:rgb_color_space, %{white: white} = opts}),
```
carson-katri, line 49: This naming doesn't feel right, since the color space is an option and this doesn't clearly show which color space it is. Could a color instead be represent with:

```ex
{:s_rgb, %{ red: 1, green: 0, blue: 0 }} # specify the sRGB space manually
{:display_p3, %{ red: 0, green: 1, blue: 1 }} # specify the displayP3 space manually
%{ red: 1, green: 0, blue: 0 } # defaults to be sRGB
```

Sources/LiveViewNative/Modifiers/ForegroundStyleModifier.swift
```diff
@@ -0,0 +1,37 @@
...
+
+import SwiftUI
+
+struct ForegroundStyleModifier: ViewModifier, Decodable {
+    private let primary: WrappedShapeStyle?
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.primary = try container.decodeIfPresent(WrappedShapeStyle.self, forKey: .primary)
```
carson-katri, line 16: You can make `primary` non-optional to avoid the force unwrap in `body(content:)`
```suggestion
        self.primary = try container.decode(WrappedShapeStyle.self, forKey: .primary)
```

Sources/LiveViewNative/Utils/WrappedShapeStyle.swift
supernintendo: Thanks, adding an extension for `AnyShapeStyle` seems to be the way to go! Fixed 🙂

lib/live_view_native_swift_ui/types/color.ex
supernintendo: This is fixed. I went with the suggestion of passing the color space as the first element of the tuple, since the tuple form is only used in this case.

Sources/LiveViewNative/Modifiers/ForegroundStyleModifier.swift
supernintendo: Good call. Fixed

Sources/LiveViewNative/Registries/BuiltinRegistry.swift
```diff
@@ -173,6 +173,8 @@ struct BuiltinRegistry: BuiltinRegistryProtocol {
     
     enum ModifierType: String {
         case fontWeight = "font_weight"
+        case backgroundStyle = "background_style"
```
shadowfacts, line 176: Can you put these in alphabetical order?

Sources/LiveViewNative/Views/Drawing and Graphics/Color.swift
```diff
@@ -46,23 +46,65 @@ struct Color<R: RootRegistry>: View {
...
         }
     }
 }
 
-extension SwiftUI.Color {
+extension SwiftUI.Color: Decodable {
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        
+        if let rgbColorSpace = try container.decode(SwiftUI.Color.RGBColorSpace?.self, forKey: .rgbColorSpace) {
```
shadowfacts, line 83: For all of these, I think you want `decodeIfPresent` rather than decoding an optional type—the absence of the key should be permitted, but currently this would throw if the key's not present.

Sources/LiveViewNative/Views/Drawing and Graphics/Gradient.swift
```diff
@@ -0,0 +1,29 @@
...
+import SwiftUI
+
+extension Gradient: Decodable {
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        var gradient: Gradient?
+
+        if let colors = try container.decode([SwiftUI.Color]?.self, forKey: .colors) {
+            gradient = Gradient(colors: colors)
+        } else if let stops = try container.decode([Gradient.Stop]?.self, forKey: .stops) {
```
shadowfacts, line 16: `decodeIfPresent` here too

shadowfacts (changes requested):
I left a couple minor suggestions, but overall this looks good :+1:

Sources/LiveViewNative/Views/Drawing and Graphics/Color.swift
supernintendo: @shadowfacts Thanks, that makes sense. This should be fixed in b17cc69

supernintendo:
Thanks @shadowfacts! I just pushed up changes for both of those suggestions.

shadowfacts (approved):

Issue #114, opened 2023-01-03
SwiftUI -> Layout Containers -> Collection containers: TableRow


AZholtkevych:
Doc: https://developer.apple.com/documentation/swiftui/tablerow

Issue #113, opened 2023-01-03
SwiftUI -> Layout Containers -> Collection containers: TableColumn


AZholtkevych:
Doc: https://developer.apple.com/documentation/swiftui/tablecolumn

PR #243, opened 2023-02-21
Add Table

### LINKED ISSUE #112: SwiftUI -> Layout Containers -> Collection containers: Table
Doc: https://developer.apple.com/documentation/swiftui/table

carson-katri:
Closes #112, #114, #113.

The markup is modeled after HTML table markup:

```heex
<table>
  <table:columns>
    <table-column width="100">Column A</table-column>
    <table-column>B</table-column>
    <table-column min-width="100" max-width="500" ideal-width="50">Column C</table-column>
  </table:columns>
  <table:rows>
    <table-row id="0">
      <text>A value</text>
      <text>B value</text>
    </table-row>
    <table-row id="1">
      A2 value
      <text>B2 value</text>
      <progress-view />
    </table-row>
    <table-row id="2">
      <rectangle fill-color="system-red" />
      <rectangle fill-color="system-green" />
      <rectangle fill-color="system-blue" />
    </table-row>
  </table:rows>
</table>
```

Each `<table-row>` must have an `id` value.

> **Note** The table content builder does not currently support `ForEach`, so for now it is limited to 10 columns.

https://user-images.githubusercontent.com/13581484/220427312-7c449c02-68d1-406c-9199-512e49d03359.mp4

Simulator Screen Shot - iPad Pro (11-inch) (4th generation) - 2023-02-21 at 13 34 17

### `<td>`
Children of the row are mapped to the column at the same index. Raw text must be nested in some way, you can use the `<text>` view or `<group>` or some other container to delineate.

Alternatively, an element like `<td>` in HTML could be added.

### Nesting Syntax
Another possible syntax would be to nest the columns in each row, as implemented in e5cfb7f.
```html
<table>
  <table-row id="0">
    <table-column id="Column A">A value</table-column>
    <table-column id="B">B value</table-column>
  </table-row>
  <table-row id="1">
    <table-column id="Column A">A2 value</table-column>
    <table-column id="B">B2 value</table-column>
    <table-column id="Column C"><progress-view /></table-column>
  </table-row>
  <table-row id="2">
    <table-column id="Column A"><rectangle fill-color="system-red" /></table-column>
    <table-column id="B"><rectangle fill-color="system-green" /></table-column>
    <table-column id="Column C"><rectangle fill-color="system-blue" /></table-column>
  </table-row>
</table>
```
However, this makes it more difficult to add modifiers to the columns (like `width`) and requires the column ID to be copied exactly for each row.


Sources/LiveViewNative/Views/Layout Containers/Collection Containers/Table.swift
```diff
@@ -0,0 +1,157 @@
...
+//  Created by Carson Katri on 2/21/23.
+//
+
+#if !os(watchOS)
+import SwiftUI
+
+struct Table<R: CustomRegistry>: View {
+    @ObservedElement private var element: ElementNode
+    private let context: LiveContext<R>
+    @Environment(\.self) private var environment
```
shadowfacts, line 14: What is this used for? It will, I believe, cause the view to be re-rendered every time any environment value changes, so I'd really like to avoid it.

shadowfacts (changes requested):

carson-katri:
Selection is now supported with LiveBindings.

https://user-images.githubusercontent.com/13581484/220445828-88c4a76d-9b50-4c98-894c-3d23a85129b8.mov

carson-katri:
Sorting is now implemented using a LiveBinding. It passes the `id` of the column to sort with, and the `order` which is `true` for forward, `false` for reverse.

Here I sort in the loop over `@raw_episodes`. It would probably be more efficient to store the sorted episodes in assigns.

```html
<table table-style="inset" sort-order="sort">
  <table:columns>
    <table-column id="number" width="50">Episode</table-column>
    <table-column id="name" ideal-width="100">Name</table-column>
    <table-column id="summary" min-width="100" max-width="500" ideal-width="50">Description</table-column>
  </table:columns>
  <table:rows>
    <%= for episode <- Enum.sort_by(@raw_episodes, fn e -> e[Map.get(hd(@sort), "id")] end, (if Map.get(hd(@sort), "order"), do: &</2, else: &>/2)) do %>
      <table-row id={episode["id"] |> Integer.to_string}>
        <text>
          S<%= episode["season"] %> E<%= episode["number"] %>
        </text>
        <text><%= episode["name"] %></text>
        <group>
          <%= if episode["summary"] != nil do %>
            <text><%= episode["summary"] |> String.replace("<p>", "") |> String.replace("</p>", "") %></text>
          <% end %>
        </group>
      </table-row>
    <% end %>
  </table:rows>
</table>
```

https://user-images.githubusercontent.com/13581484/220465124-b6ddb037-8659-4796-9a4d-4f695a3b1e5a.mov

bcardarella:
How do the IDs work? Are they scoped to just that table or will they leak to other views if they're not unique?

carson-katri:
@bcardarella The `<table>` looks up the IDs of the `<table-column>` elements that are direct children of its `<table:columns>` block. That ID is optional and is used to report to the backend which column to sort by. If you don't specify an ID it will pass the index of the column instead. You could have a separate assign for sorting separate tables, and they won't overlap at all unless you pass the same `sort-order` binding to both.

The ID for each `<table-row>` is required because the rows of a table need to be identifiable.

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
```diff
@@ -162,7 +162,6 @@ public struct LiveBinding<Value: Codable> {
         nonmutating set {
             // update the local value
             data.mode = .local(newValue)
-            data.objectWillChange.send()
```
shadowfacts, line 165: This was necessary to get the menu-style picker to update when a menu option is tapped. Is it causing a problem elsewhere?

Sources/LiveViewNative/Views/Layout Containers/Collection Containers/Table.swift
```diff
@@ -0,0 +1,320 @@
...
+            .filter { $0.tag == "table-column" }
+        return columnElements.enumerated().map { item in
+            TableColumn(item.element.innerText(), sortUsing: ColumnSort(id: item.element.attributeValue(for: "id") ?? String(item.offset), order: .forward)) { (row: Row) in
+                let rowChildren = row.element.children()
+                if rowChildren.indices.contains(item.offset) {
+                    context.coordinator.builder.fromNodes(
+                        [rowChildren[item.offset]],
+                        context: context
+                    )
+                    .environment(\.coordinatorEnvironment, coordinatorEnvironment)
```
shadowfacts, line 273: The coordinator environment values should be applied through `fromNodes`, why is it necessary to override them with the table's here?

shadowfacts (changes requested):

Sources/LiveViewNative/Views/Layout Containers/Collection Containers/Table.swift
carson-katri: The `element` environment value is set in `fromNodes`, but the `coordinatorEnvironment` only seems to be set in the `NavStackEntryView`. It was crashing in `@ObservedElement` because it was missing that environment value when I scrolled down the table.

Sources/LiveViewNative/Property Wrappers/LiveBinding.swift
carson-katri: I was receiving runtime warnings about modifying state during a view update from this line at one point, but I can't replicate it anymore.

shadowfacts: Hmm, that warning would make sense, but I'm not able to reproduce it either. We'll have to keep an eye out for it :/

Sources/LiveViewNative/Views/Layout Containers/Collection Containers/Table.swift
shadowfacts: Ah, weird that it's  not propagating through the table. Smells like a SwiftUI bug

shadowfacts (approved):

Sources/LiveViewNative/Views/Layout Containers/Collection Containers/Table.swift
carson-katri: Probably, yeah.

Issue #125, opened 2023-01-03
SwiftUI -> Layout Containers -> Presentation containers: TabView


AZholtkevych:
Doc: https://developer.apple.com/documentation/swiftui/tabview

PR #249, opened 2023-02-23
Add `TabView`

### LINKED ISSUE #39: TabView
Tracking issue/discussion for an element representing SwiftUI's TabView.

## Design Considerations
Would you want tabs to be separate pages (similar to how hierarchical navigation works) or have the view tree for each page specified together?

Linked:
```html
<tabview>
	<tab title="Tab One" target="/one" />
	<tab title="Tab Two" target="/two" />
</tabview>
```

Embedded:
```html
<tabview>
	<tab title="Tab One">
		<text>One</text>
	</tab>
	<tab title="Tab Two">
		<text>Two</text>
	</tab>
	<!-- etc. -->
</tabview>
```

## Implementation
Initially at least, it would likely be limited to client-side navigation (no switching tabs from the backend). Backend navigation is blocked by the ability to create two-way bindings between the client and the backend LiveView, as I discussed in #34.

In iOS, the convention is navigation happens _within_ the tab, but the current navigation implementation expects to be at the outermost level of the view tree.
- There would need to be separate navigation coordinators for each tab (since there's 1 per navigation stack)
- The configuration needs to separate whether navigation is permitted from at what level of the view tree navigation happens

carson-katri:
Closes #39 

I implemented this as a new `NavigationMode`. It can be used like so:

```swift
config.navigationMode = .tabView(tabs: [
  .init(label: Label("Received", systemImage: "tray.and.arrow.down.fill"), url: URL(string: "http://localhost:4000")!),
  .init(label: Label("Account", systemImage: "person.crop.circle.fill"), url: URL(string: "http://localhost:4000/new")!),
])
```

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
@@ -1,11 +1,11 @@
 //
 //  LiveSessionConfiguration.swift
-// LiveViewNative
+//  LiveViewNative
 //
 //  Created by Shadowfacts on 3/30/22.
 //
 
-import Foundation
+import SwiftUI
 
 /// An object that configures the behavior of a ``LiveSessionCoordinator``.
 public struct LiveSessionConfiguration {
@@ -37,6 +37,20 @@ public struct LiveSessionConfiguration {
         case enabled
         /// Navigation is fully enabled and uses a `NavigationSplitView` for the UI.
         case splitView
+        /// Navigation is fully enabled and uses a `TabView` to switch between routes.
+        case tabView(tabs: [Tab])
+        
+        public struct Tab: Identifiable {
+            let label: SwiftUI.Label<SwiftUI.Text, SwiftUI.Image>
+            let url: URL
+            
+            public var id: URL { url }
+            
+            public init(label: SwiftUI.Label<SwiftUI.Text, SwiftUI.Image>, url: URL) {
+                self.label = label
+                self.url = url
+            }
+        }
         
         var permitsRedirects: Bool {
             switch self {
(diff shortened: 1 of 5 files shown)


carson-katri:
Here you can see switching tabs and navigation within a tab.

https://user-images.githubusercontent.com/13581484/220963571-6daf029f-c689-4fcb-b4b9-e515a6d50970.mp4

bcardarella:
From LiveView's perspective is this being treated as a Live nav event or will this force LV to form a new connection?

carson-katri:
It does a reconnect

bcardarella:
Implementation of this view is interesting as this specific view is what sparked the idea of LiveView Native years ago when we were building the DriveIn app. So now that this is done we've come full circle 😃

carson-katri:
Just noticed there were two issues for TabView, #39 and #125. In #39 it is represented as markup, although I'm not sure it can be implemented that way since navigation needs to be inside the TabView...

bcardarella:
We can close #39 as it was an earlier exploration and likely uninformed with the growth of this lib. I'll close

carson-katri:
I think Sockets will still be retained by these lines, I think `delegate` should be `weak`.

https://github.com/davidstump/SwiftPhoenixClient/blob/7c3aaa51efc489f61589f77918b491f58299b07a/Sources/SwiftPhoenixClient/Socket.swift#L252

https://github.com/davidstump/SwiftPhoenixClient/blob/7c3aaa51efc489f61589f77918b491f58299b07a/Sources/SwiftPhoenixClient/PhoenixTransport.swift#LL189C15-L189C15

shadowfacts (approved):
I think this is a good approach for the tab view, since it moves the navigation context to inside each tab. There are probably still more places where the channel/socket are leaked, but I don't think it's worth fretting over now, since all of that code will get rewritten with the switch to core-managed channels.

PR #251, opened 2023-02-23
Add scroll-position attribute

### LINKED ISSUE #121: SwiftUI -> Layout Containers -> Scroll views: ScrollViewReader
Doc: https://developer.apple.com/documentation/swiftui/scrollviewreader

carson-katri:
Closes #121 

This is intended to cover `ScrollViewReader`'s use cases. Pass the `scroll-position` attribute to `<scroll-view>` and optionally a unit point for the anchor. Any time the value changes, the View with that ID will be scrolled into frame.

Currently, no animations are provided. These could be enabled later with a `scroll-position-animation` attribute once we have an `Animation` type in the Elixir library.

I have also updated the builder so that any time a View has an `id` attribute it is given an `.id(_:)` modifier. This could also be refactored to be a normal modifier for more control over where it applies in the stack.

```html
<z-stack alignment="bottom">
  <scroll-view scroll-position={@scroll_position} scroll-position-anchor={%LiveViewNativeSwiftUi.Types.UnitPoint{ x: 0, y: 0 } |> Jason.encode!}>
    <button phx-click="scroll_to" phx-value-id="100">Scroll to Bottom</button>
    <vstack spacing="0">
      <%= for i <- 0..100 do %>
        <color red={i / 100} green={1 - (i / 100)} blue={0.5} id={i |> Integer.to_string} modifiers={@native |> frame(height: 20)} />
      <% end %>
    </vstack>
    <button phx-click="scroll_to" phx-value-id="0">Scroll to Top</button>
  </scroll-view>
  <phx-form id="scroll_form" phx-submit="scroll_to">
    <group-box modifiers={@native |> padding(all: 16)}>
        <group-box:label><text-field name="id" format="number" placeholder="Position" /></group-box:label>
        <group-box:content><phx-submit-button button-style="bordered-prominent">Scroll</phx-submit-button></group-box:content>
    </group-box>
  </phx-form>
</z-stack>
```

https://user-images.githubusercontent.com/13581484/221023557-e7bbe167-616a-4810-842c-23dbdfa387af.mp4

### DIFF
--- a/Sources/LiveViewNative/Utils/UnitPoint.swift
+++ b/Sources/LiveViewNative/Utils/UnitPoint.swift
@@ -10,8 +10,7 @@ import LiveViewNativeCore
 import RegexBuilder
 
 extension UnitPoint: Decodable, AttributeDecodable {
-    public init(from attribute: LiveViewNativeCore.Attribute?) throws {
-        guard let value = attribute?.value else { throw AttributeDecodingError.missingAttribute(Self.self) }
+    public init(from value: String) throws {
         switch value {
         case "zero":
             self = .zero
@@ -34,9 +33,17 @@ extension UnitPoint: Decodable, AttributeDecodable {
         case "bottom-trailing":
             self = .bottomTrailing
         default:
+            let doublePattern = Regex {
+                Optionally("-")
+                OneOrMore(.digit)
+                Optionally {
+                    "."
+                    OneOrMore(.digit)
+                }
+            }
             let pattern = Regex {
                 Capture {
-                    OneOrMore(.digit)
+                    doublePattern
                 } transform: { Double($0) }
                 OneOrMore {
                     ChoiceOf {
@@ -45,7 +52,7 @@ extension UnitPoint: Decodable, AttributeDecodable {
                     }
                 }
                 Capture {
-                    OneOrMore(.digit)
+                    doublePattern
                 } transform: { Double($0) }
             }
             .anchorsMatchLineEndings()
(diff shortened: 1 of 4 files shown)


Sources/LiveViewNative/Views/Layout Containers/Scroll Views/ScrollView.swift
```diff
@@ -16,11 +16,28 @@ struct ScrollView<R: CustomRegistry>: View {
...
+                proxy.scrollTo(newValue, anchor: scrollPositionAnchor)
+            }
         }
     }
+    
+    private var scrollPositionAnchor: UnitPoint? {
+        element.attributeValue(for: "scroll-position-anchor")?
+            .data(using: .utf8)
+            .flatMap { try? JSONDecoder().decode(UnitPoint.self, from: $0) }
+    }
```
shadowfacts, line 42: Can you update this to use `@Attribute`?

shadowfacts (changes requested):
I request one slight change, plus conflict fixes, but I think this is a reasonable approach. It's a little odd that the id modifier gets a special attribute. It's not used that commonly in regular SwiftUI code, in my experience, but I agree that typing out the verbose modifiers syntax for it is excessive. I don't have a preference either way.

carson-katri:
I updated `UnitPoint` to have an `AttributeDecodable` implementation that allows values like this:
```
top-leading
center
0.5,0.5
0 1
0.25 0
```

Sources/LiveViewNative/Utils/UnitPoint.swift
```diff
@@ -6,8 +6,58 @@
...
+                    ChoiceOf {
+                        ","
+                        OneOrMore(.whitespace)
+                    }
+                }
+                Capture {
+                    OneOrMore(.digit)
+                } transform: { Double($0) }
+            }
+            .anchorsMatchLineEndings()
```
shadowfacts, line 51: Does the `.digit` character class properly handle decimals or is it just `[0-9]`?

carson-katri: You're right, I'll need to rework that

shadowfacts (approved):

Issue #178, opened 2023-02-11
Extracting PhxForm and PhxSubmitButton to their own library


bcardarella:
I'd like to get [PhxForm](https://github.com/liveviewnative/liveview-client-swiftui/blob/main/Sources/LiveViewNative/Views/PhxForm.swift) and [PhxSubmitButton](https://github.com/liveviewnative/liveview-client-swiftui/blob/main/Sources/LiveViewNative/Views/PhxSubmitButton.swift) extracted out into a library. These two views are not part of SwiftUI but enable necessary behavior for building LiveView Native apps.

These should be extracted into a new library: `liveview-native-swiftui-form`

AZholtkevych:
Brian have to make repository for LVN team

bcardarella:
Done: https://github.com/liveviewnative/liveview-native-swiftui-form

AZholtkevych:
Thanks a lot, FYI @carson-katri @shadowfacts @supernintendo

bcardarella:
@AZholtkevych this job isn't completed, just that the repo was created. The work of extracting still needs to happen.

AZholtkevych:
Yes, that's why I've reopened. Closed accidentially

carson-katri:
Would we need a way of specifying multiple registries for this to be extracted?

shadowfacts:
Yes, my plan is to have a registry type to let you compose registries that looks something like:

```swift
struct MultiRegistry<First: CustomRegistry, Second: CustomRegistry>: CustomRegistry {
	enum TagName {
		case first(First.TagName)
		case second(Second.TagName)
		init?(rawValue: String) {
			if let name = First.TagName(rawValue: rawValue) {
				self = .first(name)
			} ...
		}
	}

	static func lookup(_ tag: TagName, ...) -> some View {
		switch tag {
			case .first(let name):
				First.lookup(name, ...)
			...
		}
	}
}
```

carson-katri:
That's a good solution. If I'm imagining its use right, you may want to consider including a simpler form of creating it. Nesting them could get fairly clunky when you have many addon packages:
```swift
MultiRegistry<
  MyAppRegistry,
  MultiRegistry<
    FormRegistry,
    MultiRegistry<
      ChartsRegistry,
      AVKitRegistry
    >
  >
>
```
Maybe providing something like this?
```swift
struct MyAggregateRegistry: AggregateRegistry {
  static var registries: some CustomRegistry.Type {
[...]

PR #252, opened 2023-02-24
Add aggregate registry

shadowfacts:
Allows composing multiple custom registry types, in preparation for the form components to be extracted out.

Usage:
```swift
struct AppRegistries: AggregateRegistry {
    typealias Registries = Registry3<
        MyRegistry<Self>,
        OtherRegistry<Self>,
        ThirdRegistry<Self>
    >
}
struct MyRegistry<Root: RootRegistry>: CustomRegistry {
    // ...
}
// ...

struct ContentView: View {
    @State private var session = LiveSessionCoordinator<AppRegistries>(URL(string: "http://localhost:4000")!)
    // ...
}
```

This also introduces the concept of a "root" registry, which is the registry type that's passed in to the LiveSessionCoordinator. Registries that can be composed need to be generic over the root type, as shown above, because a view created by one sub-registry needs to be able to create views from elements that are part of other sub-registries.

Unfortunately the result builder syntax proposed in #178 is not possible (inferring the associated type from the opaque result type just doesn't work, so you have to spell out the typealias anyways. I suspect this is a compiler bug).

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveNavigationEntry.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveNavigationEntry.swift
@@ -7,7 +7,7 @@
 
 import Foundation
 
-struct LiveNavigationEntry<R: CustomRegistry>: Hashable {
+struct LiveNavigationEntry<R: RootRegistry>: Hashable {
     let url: URL
     let coordinator: LiveViewCoordinator<R>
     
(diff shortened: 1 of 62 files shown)


carson-katri:
This may be the bug that's being run into here: https://github.com/apple/swift/issues/60038

carson-katri (approved):
LGTM 👍

Issue #250, opened 2023-02-23
[RFC] Declarative API for Custom Elements


carson-katri:
Currently, users can declare their own elements via the `CustomRegistry`. However, the API could be more SwiftUI-like and typesafe by using property wrappers for the attributes:

```swift
struct MyCustomElement: View {
  @Attribute("min") private var minimum: Double // crashes if this is not provided with a helpful message
  @Attribute("max") private var maximum: Double?
  @Attribute("measurement") private var measurement // default to `String?`

  @LiveContext<MyRegistry> private var liveContext

  var body: some View {
    MyCustomView(bounds: minimum...(maximum ?? 1), measurement: measurement) {
      liveContext.buildChildren(name: "label", includeDefault: true) // the `LiveContext` wrapper knows about the element, so this is simplified.
    }
  }
}
```

The same View today would be written like this:

```swift
struct MyCustomElement: View {
  @ObservedElement private var element: ElementNode
  let context: LiveContext<MyRegistry>

  var body: some View {
    let minimum = element.attributeValue(for: "min").flatMap(Double.init)!
    let maximum = element.attributeValue(for: "max").flatMap(Double.init)
    MyCustomView(bounds: minimum...(maximum ?? 1), measurement: element.attributeValue(for: "measurement")) {
      context.buildChildren(of: element, withTagName: "label", namespace: "my-custom-element", includeDefaultSlot: true)
    }
  }
}
```

This works well, but doesn't make it immediately obvious from looking at the definition which attributes are available and what their types are. Using property wrappers, the context can be environment-aware and automatically use the correct `ElementNode` when calling its methods.

This would mostly be additive, with the `ObservedElement` wrapper being used in the other wrappers. The live context would need to be passed down through the environment.

Internal element definitions could use this as well.

bcardarella:
I like this idea. But my Swift-fu is pretty weak so I'll let @shadowfacts weigh in

shadowfacts:
I am wholeheartedly in favor of this. The API is already moving in this direction with ObservedElement/FormState/LiveBinding, and I think adding attributes is a natural extension of it.

I'm going to do an initial pass on the context aspect, since the multi-registry changes for #178 mean the context can't be passed through the CustomRegistry methods.

Some API design suggestions for whoever ends up implementing this:
- There should be a new protocol that types that are decodable from attribute values conform to.
[...]

PR #253, opened 2023-02-28
Add @Attribute property wrapper

carson-katri:
As discussed in #250

This adds the `@Attribute` property wrapper and an `AttributeDecodable` protocol. I incorporated its usage into a few Views to start, but this can be expanded to other Views if everything checks out.

### DIFF
--- a/Sources/LiveViewNative/Property Wrappers/Attribute.swift
+++ b/Sources/LiveViewNative/Property Wrappers/Attribute.swift
@@ -8,6 +8,29 @@
 import SwiftUI
 import LiveViewNativeCore
 
+/// A property wrapper that decodes an attribute of an element.
+///
+/// Specify the attribute name and an optional default value. Then use the value in the body of the View.
+/// ```swift
+/// struct MyElement: View {
+///     @Attribute("count") private var count: Double = 0
+///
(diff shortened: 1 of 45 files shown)


Sources/LiveViewNative/Property Wrappers/Attribute.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    public init(wrappedValue: T? = nil, _ name: AttributeName) {
+        self.name = name
+        self.defaultValue = wrappedValue
+    }
+    
+    public init(wrappedValue: T? = nil, _ name: String) {
+        self.name = AttributeName(name: name)
+        self.defaultValue = wrappedValue
+    }
```
shadowfacts, line 25: I don't think this initializer is necessary, since AttributeName conforms to ExpressibleByStringLiteral

shadowfacts (changes requested):

carson-katri:
I added a `transform` argument (similar to what Swift Argument Parser has) for cases where the type doesn't conform to `AttributeDecodable`, or to provide a more custom decoding process for an attribute.

shadowfacts (approved):

PR #468, opened 2023-03-08
Add MultiDatePicker

### LINKED ISSUE #74: View: MultiDatePicker
Doc: https://developer.apple.com/documentation/swiftui/multidatepicker

shadowfacts:
Closes #74

Note that this element isn't a form control, since it has a collection of values, which I think is more than can reasonably be encoded as a string.

```html
<multi-date-picker value-binding="dates" start="2023-01-01" end="2023-01-31">
  <text>Dates</text>
</multi-date-picker>
```

https://user-images.githubusercontent.com/7091588/223870309-7d1fad3d-219d-4c84-8ae5-66d90c82b99c.mp4

### DIFF
--- a/Sources/LiveViewNative/Views/Controls and Indicators/Pickers/MultiDatePicker.swift
+++ b/Sources/LiveViewNative/Views/Controls and Indicators/Pickers/MultiDatePicker.swift
@@ -1,0 +1,71 @@
+//
+//  MultiDatePicker.swift
+//  LiveViewNative
+//
+//  Created by Shadowfacts on 3/8/23.
+//
+
+#if os(iOS)
+import SwiftUI
+
+struct MultiDatePicker<R: RootRegistry>: View {
+    private let context: LiveContext<R>
+    @ObservedElement private var element
+    @Attribute("start") private var start: Date?
+    @Attribute("end") private var end: Date?
+    @LiveBinding(attribute: "value-binding") private var dates: Set<SelectedDate> = []
+    
+    private var dateComponents: Binding<Set<DateComponents>> {
+        Binding {
+            Set(self.dates.map(\.dateComponents))
+        } set: {
+            self.dates = Set($0.map(SelectedDate.init(dateComponents:)))
+        }
+    }
+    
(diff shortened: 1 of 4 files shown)


bcardarella:
In your example you have `<text>Dates</text>` embedded but I don't see that on the demo video. Does it get hidden?

Sources/LiveViewNative/Views/Controls and Indicators/Pickers/MultiDatePicker.swift
```diff
@@ -0,0 +1,83 @@
...
+import SwiftUI
+
+struct MultiDatePicker<R: RootRegistry>: View {
+    private let context: LiveContext<R>
+    @ObservedElement private var element
+    @Attribute("start") private var start: Date?
+    @Attribute("end") private var end: Date?
+    @LiveBinding(attribute: "value-binding") private var dates: [SelectedDate] = []
+    @State private var dateComponents = Set<DateComponents>()
+    @State private var skipSettingDateComponents = false
```
carson-katri, line 18: I think this could be simplified by using a `Set` for the `dates` binding:
```suggestion
    @LiveBinding(attribute: "value-binding") private var dates: Set<SelectedDate> = []
    private var dateComponents: Binding<Set<DateComponents>> {
        Binding {
            Set(self.dates.map(\.dateComponents))
        } set: {
            self.dates = Set($0.map(SelectedDate.init(dateComponents:)))
        }
    }
```
Then you can get rid of the `onChange` modifiers.

carson-katri (changes requested):

Sources/LiveViewNative/Views/Controls and Indicators/Pickers/MultiDatePicker.swift
shadowfacts: Huh, it's weird that deselection works with a set-backed LiveBinding but not when I tried with an array, even though though converting to/from Set<DateComponents> was exactly the same.

carson-katri (approved):

PR #470, opened 2023-03-09
Generate documentation for internal and SPI types

carson-katri:
This allows documentation to be written directly on the View structs instead of in separate Markdown files.

I think this approach will make the documentation easier to maintain. However, it could pollute the documentation with types users don't care about. There is a [`@_documentation`](https://github.com/apple/swift/pull/60242) attribute coming in Swift 5.8, which could make this opt-in for View types (mark the internal types with `@_documentation(visibility: public)`).


shadowfacts:
I think moving element docs back to the corresponding view types is the right approach. But, I think it's worth waiting for Swift 5.8 and `@_documentation` to be available, rather than including all of the various internal types/methods/properties in the generated docs. Beta 3 of the current release cycle—which includes Swift 5.8—came out this week, so it hopefully shouldn't be too much longer until it's out of beta and release-Xcode supports it.

carson-katri:
I made a few tweaks to the site theme and main page (included the README contents there).

I also adjusted the `Button` docs to use a common scheme (similar to that present in the current Markdown files) that could be used for all elements to display the Attributes/Events/Children they support.

```markdown
    `<ElementName>`, short description

    Overview, potentially with some examples.

    ```html
    <ElementName attribute="value">
        <ElementName:label>
            Label Content
        </ElementName:label>
    </ElementName>
    ```

    ## Attributes
    * `disabled`
        * Boolean attribute that indicates if the element is tappable.
    * `element-style`
        * The style to apply to this element.

    ## Events
    * `phx-click`
        * Event triggered when expanded.

    ## Children
    * `label`
        * Displayed on the element before being tapped.
    * `content`
        * Elements to display when expanded.
```

It would be nice if Elixir was highlighted. I think it should be possible, since `swift-docc-render` uses `highlight.js`, so I'll have to keep looking into that.

<img width="1661" alt="Screenshot 2023-03-14 at 3 36 12 PM" src="https://user-images.githubusercontent.com/13581484/225117673-ae36aa08-dd2d-45e8-8a2a-897238b2ec6a.png">

<img width="1661" alt="Screenshot 2023-03-14 at 3 36 47 PM" src="https://user-images.githubusercontent.com/13581484/225117689-13e274e5-ef39-4523-bab6-eb53d22933d7.png">

carson-katri:
Ok, I got Elixir syntax highlighting enabled. If we do want to include this, I'm not sure if this should just be a fork of `swift-docc-render`, or the way I did it now which is to modify the the `syntax-highlight.js` script with `sed` and build the project.

Screenshot 2023-03-14 at 6 33 38 PM

bcardarella:
Is this something we could request upstream to be enabled or a configuration so we don't need to manage a fork?

carson-katri:
Haha, just edited my comment to note that. They might accept that, so I could look into opening a PR for that if you want.

bcardarella:
yes, I think that is the best option if they accept

carson-katri:
I've opened a PR to make the highlight.js languages configurable: https://github.com/apple/swift-docc-render/pull/551

shadowfacts:
I think we should forego Elixir highlighting for the moment—it's no different than the current docs, and it doesn't add the huge complexity of node & webpack to the build process. Hopefully your PR is accepted, but if not, we can revisit Elixir highlighting.

carson-katri:
Couple more changes, I think this structure could work well. Automatically during the build process, the name of the symbols are changed to be in a tag form, for example `Button` -> `<Button>`. A link to the SwiftUI Documentation is automatically added to the bottom of each element's page.

`@Attribute` and `@Event` symbols are automatically transformed to their name as well, so `buttonStyle` -> `button-style`, `click` -> `phx-click`, etc. This also makes it possible to document the attributes/events with doc comments and link to them instead of writing descriptions on the element.

This is done using a Swift Package plugin, which is called from the `build_docs` script.

Screenshot 2023-03-15 at 5 56 30 PM

Screenshot 2023-03-15 at 6 04 51 PM

shadowfacts (approved):

PR #717, opened 2023-03-21
Conform Int to AttributeDecodable

carson-katri:
This is likely to be a common attribute type.


shadowfacts (approved):
Looks good, but can you add Int to the list of conforming types in the docs?

PR #722, opened 2023-03-23
Document Image

### LINKED ISSUE #338: Documentation: SwiftUI -> User Interface Element -> Images: Image
As a PO I'd like to have documentation for the [Image](https://github.com/liveviewnative/liveview-client-swiftui/issues/48)

shadowfacts:
Closes #338


shadowfacts:
It can't, because Image can be created outside of the view tree (for use with Text and ShareLink) but `@Attribute` assumes the element is accessible via the environment. I initially refactored Text to use `@Attribute` while documenting it, before I realized that.

Issue #183, opened 2023-02-15
Thermostat application with LVN > Add missing modifiers


AZholtkevych:
Modifiers for Thermostat app:
@supernintendo please use that link https://github.com/chrismccord/phoenix_live_view_example/blob/master/lib/demo_web/live/thermostat_live.ex
* [`Color.gradient`](https://developer.apple.com/documentation/swiftui/color/gradient)
 * [`fontWeight(_:)`](https://developer.apple.com/documentation/swiftui/view/fontweight(_:))
 * [`foregroundStyle(_:)`](https://developer.apple.com/documentation/swiftui/view/foregroundstyle(_:))
 * [`background(_:in:)`](https://developer.apple.com/documentation/swiftui/view/background(_:in:fillstyle:)-20tq5)
 * [`bold()`](https://developer.apple.com/documentation/swiftui/view/bold(_:))
 * [`aspectRatio(_:contentMode:)`](https://developer.apple.com/documentation/swiftui/view/aspectratio(_:contentmode:)-6j7xz)

supernintendo:
- [ ] `Color.gradient`
- [ ] `fontWeight(_:)` - https://github.com/liveviewnative/liveview-client-swiftui/pull/202
- [ ] `foregroundStyle(_:)`
- [ ] `background(_:in:)`
- [ ] `bold()`
- [ ] `aspectRatio(_:contentMode:)`

AZholtkevych:
Waiting for front-end part

Issue #170, opened 2023-02-08
[RFC] Nested View modifier syntax


carson-katri:
There is an open question around how best to support nested modifiers, which are very common in SwiftUI. Take the [`background(alignment:content:)`](https://developer.apple.com/documentation/swiftui/view/background(alignment:content:)) modifier for example:

```swift
Text("Hello, world!")
  .background(alignment: .center) {
    Text("This is in the background!")
  }
```

Unlike other modifiers, it takes more `View`s as one of the arguments. Here are a few possibilities for supporting this:

## Content IDs
By passing an ID string to a modifier function, we can reference the labelled content within the View. This would likely only require changes on the `liveview-client-swiftui` side.

```html
<text modifiers={@native |> background(alignment: :center, content: "my-background-content")}>
    Hello, world!
    <background:my-background-content>
        <text>This text is in the background!</text>
    </background:my-background-content>
</text>
```

The namespace maps to the modifier name, and the tag name maps to the ID passed to the modifier.

This feels analogous to this SwiftUI pattern of separating out the modifier content:

```swift
var body: some View {
  Text("Hello, world!")
    .background { myBackgroundContent }
}

@ViewBuilder
var myBackgroundContent: some View {
  Text("This text is in the background!")
}
```

### Benefits
* Simple implementation
* Keeps the content of the modifier close to the View it modifies

### Drawbacks
* New users may confuse the string as a text argument to the background, and not notice the nested content.
  * This can be mitigated by using a custom type, and possibly a sigil to construct it: `background(content: Modifier.reference("my-background-content"))`, `background(content: ~R"my-background-content")`

## Template References
A template filename could be passed instead of an ID to content in the same template.

```html
<text modifiers={@native |> background(alignment: :center, content: "my-background-content.heex")}>
    Hello, world!
</text>

<!-- my-background-content.heex -->
<text>This text is in the background!</text>
```

This feels more analogous to having a separate `View` as the content:
```swift
Text("Hello, world!")
  .background { MyBackgroundContent() }

// MyBackgroundContent.swift
struct MyBackgroundContent: View {
  var body: some View {
    Text("This text is in the background!")
  }
}
```

### Benefits
* Allows modifiers to be split up more freely
* The string argument clearly represents a file

### Drawbacks
* Leads to many files quickly
* Likely requires backend changes to support loading the referenced template somewhere, perhaps mounting another LiveView to draw it

## Inline `~H`
[...]

PR #734, opened 2023-03-24
Add background modifier

supernintendo:
Adds support for the [background](https://developer.apple.com/documentation/swiftui/view/background(alignment:content:)) modifier and adds the ability to use namespaced child elements when decoding modifiers. This is based on @carson-katri's idea for handling nested views within modifiers via Content Entities as described [here](https://github.com/liveview-native/liveview-client-swiftui/issues/170).

```heex
<VStack id="background">
  <Spacer />
  <HStack id="example_heart">
    <Image system-name="heart.fill" modifiers={@native |> background(alignment: :center, content: :heart_bg) |> foreground_style(primary: {:color, :white})}>
      <background:heart_bg>
        <Circle modifiers={@native |> frame(width: 32, height: 32) |> foreground_style(primary: {:color, :red})} />
      </background:heart_bg>
    </Image>
  </HStack>
  <Spacer />
  <HStack id="example_liveviewnative">
    <Text modifiers={@native |> background(alignment: :center, content: :text_bg) |> foreground_style(primary: {:linear_gradient, %{gradient: {:colors, [:orange, :indigo, :purple]}, start_point: {0, 0}, end_point: {1, 1}}})}>
      liveviewnative
      <background:text_bg>
        <RoundedRectangle corner-radius="8" modifiers={@native |> frame(width: 192, height: 64)} />
      </background:text_bg>
    </Text>
  </HStack>
  <Spacer />
</VStack>
```

Screenshot 2023-03-23 at 5 58 14 PM

There are a few outstanding issues with this PR:

1. The `foreground_style` modifier influences the background content instead of the `background_style` modifier which doesn't appear to do anything. I'm not sure why this is but from what I can tell the native [backgroundStyle modifier](https://developer.apple.com/documentation/swiftui/grid/backgroundstyle(_:)) only affects nested content rendered using the `in:` constructor and not content passed to `content`.

2. There seem to be performance limitations with nested modifier views in my implementation. By rendering too many elements with background modifiers within the same view I'm able to produce a crash, although I haven't investigated this enough to know what might be causing it.

3. ~The string format for `content` when referencing namespaced child views could be a bit confusing. I'd like to tweak the Elixir modifier schema so that it accepts an atom instead, which should ameliorate this.~ Fixed in [2ccb51e](https://github.com/liveview-native/liveview-client-swiftui/pull/734/commits/2ccb51e8dde56862f6fbdee23503e2d304b682e1)

Closes https://github.com/liveview-native/liveview-client-swiftui/issues/183.


Sources/LiveViewNative/ViewTree.swift
```diff
@@ -93,14 +93,17 @@ struct ViewTreeBuilder<R: RootRegistry> {
...
     }
     
-    private func applyModifiers(encoded: String?, to view: some View, context: LiveContextStorage<R>) -> some View {
+    private func applyModifiers(encoded: String?, to view: some View, element: ElementNode, context: LiveContextStorage<R>) -> some View {
         let modifiers: [ModifierContainer<R>]
         if let encoded {
             let decoder = JSONDecoder()
+            decoder.userInfo[.liveContext!] = context
+            decoder.userInfo[.elementNode!] = element
+
```
shadowfacts, line 106: The context and element shouldn't be stored and passed through the decoder, they can be accessed from the modifier using their respective property wrappers:
```swift
struct MyModifier<R: RootRegistry>: ViewModifier, Decodable {
    @LiveContext<R> private var context
    // etc.
}
```

Sources/LiveViewNative/Registries/BuiltinRegistry.swift
```diff
@@ -16,10 +16,7 @@ protocol BuiltinRegistryProtocol {
     static func decodeModifier(_ type: ModifierType, from decoder: Decoder) throws -> BuiltinModifier
 }
 
-struct BuiltinRegistry: BuiltinRegistryProtocol {
-    
-    static let attributeDecoder = JSONDecoder()
-    
+struct BuiltinRegistry<R: RootRegistry>: BuiltinRegistryProtocol {
```
shadowfacts, line 19: This makes sense, but can you also remove the `R` generic type and the `context` parameter from the `lookup` method, for consistency?

shadowfacts (changes requested):
Something is definitely weird with how the background/foreground modifiers are interacting, though it's not obvious to me what the issue is.

In plain old SwiftUI, the following (translated directly from your example) results in just a red circle appearing, since the foreground style is applying to both the Circle and the Image.
```swift
Image(systemName: "heart.fill")
	.background(alignment: .center) {
		Circle()
			.frame(width: 32, height: 32)
	}
	.foregroundStyle(.red)
```

To get the result that you actually show, you'd need to move the foreground style _inside_ the background modifier so that it only applies to the Circle:
```swift
Image(systemName: "heart.fill")
	.background(alignment: .center) {
		Circle()
			.frame(width: 32, height: 32)
			.foregroundStyle(.red)
	}
```

carson-katri:
`Image` does not respect the `foreground_style` modifier because it uses a `foregroundColor` modifier in its body, which uses the system foreground color when set to `nil`. The example below produces the same result as the provided screenshot:

```swift
Image(systemName: "heart.fill")
            .foregroundColor(nil)
            .background(alignment: .center) {
                Circle()
                    .frame(width: 32, height: 32)
            }
            .foregroundStyle(.red)
```

https://github.com/liveview-native/liveview-client-swiftui/blob/931a824c85f17fc2e85924fde82b1d3d23cee6ad/Sources/LiveViewNative/Views/Images/Image.swift#L45

The attribute should probably be removed from `Image`.

Text has [the same problem](https://github.com/liveview-native/liveview-client-swiftui/blob/931a824c85f17fc2e85924fde82b1d3d23cee6ad/Sources/LiveViewNative/Views/Text%20Input%20and%20Output/Text.swift#L78). The `textColor`, `font`, and `fontWeight` attributes should likely be replaced with modifiers as well.

carson-katri:
RE the `background_style` modifier, this would require using the [`BackgroundStyle`](https://developer.apple.com/documentation/swiftui/shapestyle/background) shape style. Background content created in this way uses the foreground style by default.

shadowfacts:
Ah, that is indeed the case. We do probably want general modifiers for them, but, I think the `foregroundColor`, `font`, and `fontWeight` modifiers need to remain attributes on the Text view as well, since doing them through our usual modifiers architecture would break composing Texts with different styles.

I think a workaround would be having a helpers like `foregroundColorIfPresent` for Text that only applies the underlying modifier if the value is non-nil, rather than applying `.foregroundColor(nil)` as currently happens.

Sources/LiveViewNative/Modifiers/BackgroundModifier.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  BackgroundModifier.swift
+// LiveViewNative
+//
+//  Created by May Matyi on 3/17/23.
+//
+
+import SwiftUI
+
+struct BackgroundModifier<R: RootRegistry>: ViewModifier, Decodable {
```
carson-katri, line 10: Could you document the modifier?

supernintendo: @carson-katri Done!

supernintendo:
> 

I added `foregroundColorIfPresent` to fix that in [f7e7172](https://github.com/liveview-native/liveview-client-swiftui/pull/734/commits/f7e71724f2ecf6fa8f02e7610e241cf6e5449173). Let me know if there are other examples of this I should address, although maybe that's out of scope for this pull request?

carson-katri (approved):

shadowfacts (changes requested):
This looks good, just needs a slight tweak to make the docs visible

shadowfacts (approved):

PR #735, opened 2023-03-24
Add aspect ratio modifier

supernintendo:
Adds support for the [aspectRatio](https://developer.apple.com/documentation/swiftui/view/aspectratio(_:contentmode:)-771ow) modifier:

```heex
<VStack>
  <Spacer />
  <Image name="AspectTest" resizable modifiers={@native |> aspect_ratio(content_mode: "fit") |> frame(width: 300, height: 225)} />
  <Spacer />
  <Image name="AspectTest" resizable modifiers={@native |> aspect_ratio(content_mode: "fill") |> frame(width: 300, height: 225)} />
  <Spacer />
</VStack>
```

Screenshot 2023-03-23 at 8 42 56 PM

### DIFF
--- a/Sources/LiveViewNative/Modifiers/AspectRatioModifier.swift
+++ b/Sources/LiveViewNative/Modifiers/AspectRatioModifier.swift
@@ -1,0 +1,37 @@
+//
+//  AspectRatioModifier.swift
+// LiveViewNative
+//
+//  Created by May Matyi on 3/23/23.
+//
+
+import SwiftUI
+
+struct AspectRatioModifier: ViewModifier, Decodable {
+    private let aspectRatio: CGSize
+    private let contentMode: ContentMode
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.aspectRatio = try container.decode(CGSize.self, forKey: .aspectRatio)
+
+        switch try container.decode(String.self, forKey: .contentMode) {
+        case "fill":
+            self.contentMode = .fill
+        case "fit":
+            self.contentMode = .fit
+        default:
+            throw DecodingError.dataCorruptedError(forKey: .contentMode, in: container, debugDescription: "invalid value for contentMode")
+        }
+    }
+
+    func body(content: Content) -> some View {
+        content.aspectRatio(self.aspectRatio, contentMode: self.contentMode)
+    }
+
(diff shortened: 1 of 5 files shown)


Sources/LiveViewNative/Modifiers/AspectRatioModifier.swift
```diff
@@ -0,0 +1,37 @@
...
+import SwiftUI
+
+struct AspectRatioModifier: ViewModifier, Decodable {
+    private let aspectRatio: CGSize
+    private let contentMode: ContentMode
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.aspectRatio = try container.decode(CGSize.self, forKey: .aspectRatio)
```
shadowfacts, line 17: The aspect ratio should be optional, since it is in SwiftUI.

The body of the modifier will have to be slightly different, since the aspect ratio-less variant takes a `CGFloat?` not a `CGSize?`:

```swift
if let aspectRatio {
    content.aspectRatio(aspectRatio, contentMode: contentMode)
} else {
    content.aspectRatio(contentMode: contentMode)
}
```

Sources/LiveViewNative/Views/Images/Image.swift
```diff
@@ -28,6 +28,7 @@ import SwiftUI
 @_documentation(visibility: public)
 #endif
 struct Image: View {
+    @Attribute("resizable") private var resizable: Bool
```
shadowfacts, line 31: Can you add a doc comment (and the related dance for `@_documentation`) for this?

shadowfacts (changes requested):

Sources/LiveViewNative/Modifiers/AspectRatioModifier.swift
supernintendo: Oh, good catch. Fixed!

Sources/LiveViewNative/Views/Images/Image.swift
supernintendo: Done! I noticed the compiler directive from other documentation so I went ahead and copied that, but not sure if it's needed. Also feel free to suggest changes to the documentation text I wrote, not sure if there are any rules / conventions around that.

shadowfacts (approved):

PR #738, opened 2023-03-28
Document *SplitView

carson-katri:


Sources/LiveViewNative/Views/Layout Containers/Presentation Containers/HSplitView.swift
```diff
@@ -5,17 +5,33 @@
...
 //
 
-#if os(macOS)
 import SwiftUI
 
+/// Container with resizable splits between elements.
+///
+/// Each element in the container will be in a separate resizable section.
+///
+/// If the element is not flexible, the section will only be able to shrink.
```
shadowfacts, line 14: Can you clarify what flexible means? Since I don't think we're using that terminology elsewhere

shadowfacts (changes requested):

Sources/LiveViewNative/Views/Layout Containers/Presentation Containers/HSplitView.swift
carson-katri: Let me know if cf1cb5d clears it up, or if there's some other way I should describe this.

PR #775, opened 2023-04-06
Add `font` modifier

### LINKED ISSUE #217: Modifier -> SwiftUI -> Text input and output: font(_:)
Doc: https://developer.apple.com/documentation/swiftui/view/font(_:)

carson-katri:
Closes #217

### DIFF
--- a/Sources/LiveViewNative/Utils/Font.swift
+++ b/Sources/LiveViewNative/Utils/Font.swift
@@ -202,49 +202,23 @@ extension Font: Decodable {
 /// * `bold`
 /// * `heavy`
 /// * `black`
-///
-/// You can also use the numeric value for the associated weight, such as `400` for regular and `900` for `black`.
 #if swift(>=5.8)
 @_documentation(visibility: public)
 #endif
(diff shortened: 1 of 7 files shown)


Sources/LiveViewNative/Utils/Font.swift
```diff
@@ -0,0 +1,358 @@
...
+            case "medium": self = .medium
+            case "semibold": self = .semibold
+            case "bold": self = .bold
+            case "heavy": self = .heavy
+            case "black": self = .black
+            default: throw DecodingError.dataCorrupted(.init(codingPath: [], debugDescription: "Unknown weight named \(name)"))
+            }
+        } else {
+            switch try container.decode(Int.self) {
+            case 100:
```
shadowfacts, line 228: These values aren't defined by SwiftUI(?), so I'm not sure if we should permit them?

carson-katri: Good point, removed.

shadowfacts (approved):

PR #778, opened 2023-04-06
Add `text_field_style` modifier

### LINKED ISSUE #479: Modifier -> SwiftUI -> Text input and output: textFieldStyle(_:)
Doc: https://developer.apple.com/documentation/swiftui/view/textfieldstyle(_:)

- [ ] Swift implementation
- [ ] Elixir implementation

carson-katri:
Closes #479 

Should the styles be underscored instead of dashed? Currently we can set the style of the field two ways:

```html
<TextField text-field-style="rounded-border" />
<TextField modifiers={text_field_style(@native, style: :"rounded-border")} />
```

If we make it underscored for both, the atom wouldn't need quotes. Could also use a different name for the attribute vs. modifier method.

### DIFF
--- a/Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/TextFieldStyleModifier.swift
+++ b/Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/TextFieldStyleModifier.swift
@@ -38,10 +38,50 @@ struct TextFieldStyleModifier: ViewModifier, Decodable {
     }
 
     func body(content: Content) -> some View {
-        content.applyTextFieldStyle(style)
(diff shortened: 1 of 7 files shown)


shadowfacts:
I started on the other style modifiers (`button_style`, etc.), and I've been replacing the existing attributes with the modifiers. So, the values are underscored and consistent with Elixir atom naming conventions. I think that's preferable, since the style modifiers should be consistent when applied to just a single control vs a larger portion of the view tree. Thoughts?

carson-katri:
Yeah, that sounds good to me.

shadowfacts (approved):

PR #787, opened 2023-04-12
Use `<... template={...}>` for named children

carson-katri:
This changes the syntax for named children. Two main ways are supported:

1. `template={...}` attribute

This allows dynamic values to be used easily.
```html
<Text template={:label}> ... </Text>
```

2. Shorthand `#` boolean attribute

This can be preferred when the value is constant.
```html
<Text #label> ... </Text>
```

### Comparisons
Left is current syntax, right is new syntax.

<table>
<tr>
<td>

```html
<Menu>
  <Menu:label>
    Open Me
  </Menu:label>
  <Menu:content>
    <Button>Option 1</Button>
    <Button>Option 2</Button>
    <Button>Option 3</Button>
  </Menu:content>
</Menu>
```

</td>
<td>

```html
<Menu>
  <Text #label>
    Open Me
  </Text>
  <Group #content>
    <Button>Option 1</Button>
    <Button>Option 2</Button>
    <Button>Option 3</Button>
  </Group>
</Menu>
```

</td>
</tr>
</table>

Here's one case where this is very beneficial: dynamically changing what part of the content a template represents.

<table>
<tr>
<td>

```html
<Gauge>
  <Gauge:label>Value</Gauge:label>
  <%= content_tag :"Gauge:#{@min_or_max_label}" do %>
    This is either the minimum-value-label or maximum-value-label
  <% end %>
</Gauge>
```

</td>
<td>

```html
<Gauge>
  <Text #label>Value</Text>
  <Text template={@min_or_max_label}>
    This is either the minimum-value-label or maximum-value-label
  </Text>
</Gauge>
```

</td>
</tr>
</table>

According to the docs, the use of `content_tag` is discouraged. However I believe this is the only way that could be implemented with the current syntax. The new syntax makes this easy as you just pass the `id` of the template like a normal attribute.

### DIFF
--- a/Sources/LiveViewNative/Property Wrappers/LiveContext.swift
+++ b/Sources/LiveViewNative/Property Wrappers/LiveContext.swift
@@ -60,11 +60,11 @@ public struct LiveContext<R: RootRegistry>: DynamicProperty {
     
     private static func elementWithName(
         _ tagName: String,
-        namespace: String?
+        id: String
     ) -> (NodeChildrenSequence.Element) -> Bool {
         { child in
             if case let .element(element) = child.data,
-               element.namespace == namespace,
+               element.attributes.first(where: { $0.name == "id" })?.value == id,
                element.tag == tagName
             {
                 return true
@@ -75,44 +75,41 @@ public struct LiveContext<R: RootRegistry>: DynamicProperty {
     }
     
     /// Checks whether the element has any children with the given tag.
-    public func hasChild(
+    public func hasTemplate(
         of element: ElementNode,
-        withTagName tagName: String,
-        namespace: String? = nil
+        withID id: String
     ) -> Bool {
-        element.children().contains(where: Self.elementWithName(tagName, namespace: namespace))
+        element.children().contains(where: Self.elementWithName("template", id: id))
     }
     
     /// Builds a view representing only the children of the element which have the given tag name.
     ///
     /// This can be use to build views which have multiple types of children, such as how Menu takes content and a label:
     /// ```html
-    /// <Menu>
-    ///     <Menu:content>
-    ///         <Button phx-click="clicked">Hello</Button>
-    ///     </Menu:content>
-    ///     <Menu:label>
+    /// <menu>
+    ///     <template id={:content}>
+    ///         <button phx-click="clicked">Hello</button>
+    ///     </template>
+    ///     <template id={:label}>
     ///         My Menu
-    ///     </Menu:label>
-    /// </Menu>
+    ///     </template>
+    /// </menu>
     /// ```
     ///
     /// - Parameter element: The element whose children to consider.
-    /// - Parameter withTagName: The name of the tag to build children from.
-    /// - Parameter namespace: The namespace of the tag to build children from.
-    /// - Parameter includeDefaultSlot: Whether to use all un-namespaced children if there are no children with the correct tag name and namespace.
+    /// - Parameter id: The ID of the `<template>` containing the children.
+    /// - Parameter includeDefaultSlot: Whether to use all other children if there is no `<template>` with the correct id.
     public func buildChildren(
         of element: ElementNode,
-        withTagName tagName: String,
-        namespace: String? = nil,
+        withID id: String,
         includeDefaultSlot: Bool = false
     ) -> some View {
         let children = element.children()
-        let namedSlotChildren = self.children(of: element, withTagName: tagName, namespace: namespace)
+        let namedSlotChildren = children.filter(Self.elementWithName("template", id: id))
         if namedSlotChildren.isEmpty && includeDefaultSlot {
             let defaultSlotChildren = children.filter({
                 if case let .element(element) = $0.data {
-                    return element.namespace != namespace
+                    return element.tag != "template"
                 } else {
                     return true
                 }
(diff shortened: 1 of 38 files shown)


bcardarella:
Just for tracking purposes, I'm :+1: on this proposal. I would just like to ensure that on the Elixir side we can use `<t>` in addition to `<template>` which would look like this:

```html
<Menu>
  <t id={:label}>
    Open Me
  </t>
  <t id={:content}>
    <Button>Option 1</Button>
    <Button>Option 2</Button>
    <Button>Option 3</Button>
  </t>
</Menu>
```

shadowfacts:
Personally, I have a slight preference for having a `<template>` element, since it's closer to the Web Components and Vue.js style—and it indicates the same thing: that the contents will be placed somewhere deeper within the parent element's internal structure. It is less concise, but I think it adds some clarity. But that's not a strong preference, and I'm fine with the attribute style.

bcardarella:
@shadowfacts agreed, I retract my previous request

carson-katri:
Option 2 is a bit shorter in many cases (even if we don't include the shorthand), which was the main concern with `<template>`. But I also see the benefit of forcing that extra nesting level with `<template>`.

> The attribute method is more like [how WebComponents work](https://developer.mozilla.org/en-US/docs/Web/API/Element/slot) than the `<template>` element which is [similar to Vue](https://vuejs.org/guide/components/slots.html#named-slots).

Do either of you have a strong preference between these @bcardarella @supernintendo?

@Kashif-E & @KronicDeth: would one of these options be better for Jetpack Compose?

### Option 1
> Requires a `<template>` element with an `id`.
```html
<Menu>
  <template id={:label}>
    <Text>Open Me</Text>
  </template>
  <template id={:content}>
    <Button>Option 1</Button>
    <Button>Option 2</Button>
    <Button>Option 3</Button>
  </template>
</Menu>
```

### Option 2
> No need for a `<template>` *element*, as the `template` *attribute* can be applied to any element.
```html
<Menu>
  <Text #label> <!-- this # attribute is shorthand for template={:label} -->
    Open Me
  </Text>
  <Group template={:content}> <!-- use the Group view to put multiple elements in this part of the menu -->
    <Button>Option 1</Button>
    <Button>Option 2</Button>
    <Button>Option 3</Button>
  </Group>
</Menu>
```

### Equivalent SwiftUI View
```swift
Menu {
  Button("Option 1") {}
  Button("Option 2") {}
  Button("Option 3") {}
} label: {
  Text("Open Me")
}
```

Kashif-E:
@carson-katri in jetpack compose attributes or modifiers are chained together, hence, my personal preference will be option 2 because of the same reason you mentioned above

Sources/LiveViewNative/Property Wrappers/LiveContext.swift
```diff
@@ -58,14 +58,13 @@ public struct LiveContext<R: RootRegistry>: DynamicProperty {
...
-        namespace: String?
+    private static func isTemplateElement(
+        _ template: String
     ) -> (NodeChildrenSequence.Element) -> Bool {
         { child in
             if case let .element(element) = child.data,
-               element.namespace == namespace,
-               element.tag == tagName
+               element.attributes.first(where: { $0.name == "template" })?.value == template
+                || element.attributes.contains(where: { $0.name == .init(stringLiteral: "#\(template)") })
```
shadowfacts, line 67: I was under the impression that the # syntax was going to be an EEx engine transformation to the same template attribute?

carson-katri: I'm not sure how to implement it that way with the `TagEngine`. @supernintendo Is this possible and if so can you help with this?

supernintendo: @shadowfacts @carson-katri Yeah, we probably want to handle this within `TagEngine` to make it available to all platforms. The idea is to convert `<Text #label>` to `<Text template={:label}>`, correct?

shadowfacts: Yeah, that's my understanding

supernintendo: @carson-katri @shadowfacts Looking into this further, it is not currently possible to override how attributes are handled within elements from implementations of `Phoenix.LiveView.TagEngine`. `Phoenix.LiveView.TagEngine` _does_ do some parsing on attributes [here](https://github.com/phoenixframework/phoenix_live_view/blob/v0.18.18/lib/phoenix_live_view/tag_engine.ex#L934) but its not exposed through a callback in a way that we could hook into it.

I'm happy to open a PR on [phoenix_live_view](https://github.com/phoenixframework/phoenix_live_view/) to convert some of the attribute parsing functions into callbacks so we can override them, although no guarantees that will get merged. I think its probably fine to handle this within the client for now until we know we can do that. Thoughts @bcardarella?

carson-katri: I could remove the shorthand until we know whether or not LiveView will accept the changes, then we can reevaluate then.

lenileiro: This is such an elegant solution that really makes the project more approachable for new learners. it would be worth to having it in client and handle technical implementation later as @supernintendo mentioned.

carson-katri:
I've removed the `#` shorthand for now. It could be re-added if we determine it cannot be done from the backend.

shadowfacts (approved):

PR #803, opened 2023-04-14
Update tutorial section 01

murtza:
Update tutorial section 01


Sources/LiveViewNative/LiveViewNative.docc/Resources/01-01-03-config01.exs
```diff
@@ -0,0 +1,58 @@
...
+  metadata: [:request_id]
+
+# Use Jason for JSON parsing in Phoenix
+config :phoenix, :json_library, Jason
+
+config :live_view_native, :platforms, [LiveViewNativeSwiftUi.Platform]
+
+config :live_view_native, LiveViewNativeSwiftUi.Platform,
+  app_name: "LVNTutorial",
+  custom_modifiers: []
```
shadowfacts, line 54: Can you move these closer to the top of the file? The rendered tutorial doesn't do a great job indicating when there are changes below the fold, which can confuse people.

shadowfacts (changes requested):
The new tutorial resource files should be renamed to match the convention (`<tutorial number>-<section number>-<file number>`), and the ones following the new files should have their numbers adjusted, but other than that, this looks good!

murtza:
> The new tutorial resource files should be renamed to match the convention (`<tutorial number>-<section number>-<file number>`), and the ones following the new files should have their numbers adjusted, but other than that, this looks good!

What do you think about following convention `<tutorial number>-<section number>-<file number>-<file name><change number>`. This way we can show originally generated file and the changes will be highlighted.

shadowfacts:
Right now, the initial versions of files are already included (like `02-01-02-application.ex` is the initial version, and `02-01-03-application.ex` is the modified version). I'm fine with either convention, so long as we're consistent throughout the resource files.

murtza:
Ok, let's follow the existing convention `<tutorial number>-<section number>-<file number>`.

Sources/LiveViewNative/LiveViewNative.docc/Tutorials/01 Initial List.tutorial
```diff
@@ -89,13 +116,22 @@
...
+               Edit the `app.html.heex` file in the created project.
+               
+               @Comment {
+                   skip = true
+               }
+               @Code(name: "lib/lvn_tutorial_web/templates/layout/app.html.heex", file: 01-01-08-app-template.html.heex)
+           }
+
             @Step {
                 Lastly, we need to tweak the `live.html.heex` layout to remove the extraneous HTML that the native frontend doesn't need.
```
shadowfacts, line 129:
```suggestion
                Lastly, we need to tweak the `app.html.heex` layout to remove the extraneous HTML that the native frontend doesn't need.
```

shadowfacts (approved):

PR #804, opened 2023-04-14
Fix StatusBarHiddenModifier not compiling on non-iOS platforms

shadowfacts:


carson-katri:
I made a similar change in #790

PR #810, opened 2023-04-18
Use ForEach to reduce stack usage

carson-katri:
This builds on the improvements made in #809. Using `ForEach` reduces the number of calls to `f` compared to using `TupleView`.

### DIFF
--- a/Sources/LiveViewNative/ViewTree.swift
+++ b/Sources/LiveViewNative/ViewTree.swift
@@ -28,23 +28,41 @@ struct ViewTreeBuilder<R: RootRegistry> {
         case 1:
             f(e[0], c)
         case 2:
-            TupleView((f(e[0], c), f(e[1], c)))
+            ForEach(0..<2) {
+                f(e[$0], c)
+            }
         case 3:
-            TupleView((f(e[0], c), f(e[1], c), f(e[2], c)))
+            ForEach(0..<3) {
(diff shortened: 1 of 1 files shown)


carson-katri:
I tried to measure it as described in #806 with `reg read sp` and calculate the difference between `start` and the `fromNodes` call. These are the deltas:
* `main` - 610528
* `shadowfacts/fix-stack-overflow` - 115168
* `stack-foreach-optimization` - 29680

Could you confirm @shadowfacts?

shadowfacts:
I see a similar improvement. My concern is that using ForEach with an int range like this means that the the views in the ForEach are identified by their indices, which could cause state tracking issues if views are reordered. It's for that reason that the range initializer is annotated with `@_semantics("swiftui.requires_constant_range")` and the docs specifically note:
> The instance only reads the initial value of the provided `data` and doesn’t need to identify views across updates. To compute views on demand over a dynamic range, use `ForEach/init(_:id:content:)`.

carson-katri:
Would that cause problems here? The ranges are all constant, it's the `ForEach` itself that is swapped out when the count is updated.

shadowfacts:
I don't know for certain that it would cause problems, I just think it's something to be aware of. If you've tested it and there don't seem to be adverse effects, I'm good with this change.

PR #823, opened 2023-04-18
Add `text_input_auto_capitalization` modifier

### LINKED ISSUE #327: Modifier -> SwiftUI -> Text input and output: textInputAutocapitalization(_:)
Doc: https://developer.apple.com/documentation/swiftui/view/textinputautocapitalization(_:)

- [x] Swift implementation
- [x] Elixir implementation

dginsburg:
Closes #327


Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/TextInputAutocapitalizationModifier.swift
```diff
@@ -0,0 +1,65 @@
...
+/// ```
+///
+/// ## Arguments
+/// * ``autocapitalization``
+#if swift(>=5.8)
+@_documentation(visibility: public)
+#endif
+@available(iOS 16.0, watchOS 9.0, *)
+struct TextInputAutocapitalizationModifier: ViewModifier, Decodable, Equatable {
+    /// One of the capitalizing behaviors defined in the `TextInputAutocapitalization` struct or nil.
```
carson-katri, line 25: Can you list the possible values here?
```suggestion
    /// One of the capitalizing behaviors defined in the `TextInputAutocapitalization` struct or nil.
    ///
    /// Possible values:
    /// * `characters`
    /// * `sentences`
    /// * `words`
    /// * `never`
```

carson-katri (changes requested):

Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/TextInputAutocapitalizationModifier.swift
dginsburg: Of course I can and will do this moving forward. I was modeling my documentation from existing modifiers. If we want the documentation to be consistent about, well... enumerating the enums then there are some others to fix.

carson-katri: Yeah, I think its only important if we don't extend the type to be `Decodable` (in which case the cases should be listed on the extension's docs).

carson-katri (changes requested):
You could probably remove `autocapitalization` from `TextFieldProtocol` now.

carson-katri (approved):

PR #829, opened 2023-04-19
Generate BuiltinRegistry automatically

carson-katri:
This adds a plugin that runs at build time to generate the `lookup` and `decodeModifier` methods on the `BuiltinRegistry` based on the View and ViewModifier files in the package.

This should help prevent merge conflicts that commonly arise when adding views and modifiers to the registry, and goes along with the improvements made in #802.


shadowfacts (changes requested):
Can you convert it to generate the `BuiltinElement` struct from #809 for the views? Other than that, this approach looks good.

PR #834, opened 2023-04-20
Refactor and document padding modifier

carson-katri:
This refactors the padding modifier to make accessing the system padding amount possible.

```html
<Text modifiers={padding(@native)}>Use the system padding</Text>
<Text modifiers={padding(@native, edges: :horizontal, length: 10)}>Affect specific edges</Text>
<Text modifiers={padding(@native, insets: [top: 8, bottom: 16])}>Use EdgeInsets</Text>
```

### DIFF
--- a/Sources/LiveViewNative/Modifiers/Layout Adjustments Modifiers/PaddingModifier.swift
+++ b/Sources/LiveViewNative/Modifiers/Layout Adjustments Modifiers/PaddingModifier.swift
@@ -7,19 +7,85 @@
 
 import SwiftUI
 
+/// Adds space around the element.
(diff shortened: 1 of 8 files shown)


shadowfacts (approved):

carson-katri:
@shadowfacts I removed the padding modifier decoding test, since we aren't really using the `DecodeModifiersTest` extensively at the moment. Should these tests be expanded though? If so I can re-add it.

shadowfacts:
I don't think the modifier decoding tests are necessary—really we should be expanded the render tests to cover modifiers that make visible changes

PR #835, opened 2023-04-20
Optimize BuiltinRegistryGenerator

carson-katri:
Before: 49.7 seconds
After: 0.1 seconds


dginsburg (approved):
I'll code 497 times faster now. 🎉

PR #837, opened 2023-04-20
Add missing ShapeStyles

carson-katri:
Missing styles:
- [x] `HierarchicalShapeStyle`
- [x] `Material`
- [x] `AngularGradient`
- [x] `ConicGradient`
- [x] `EllipticalGradient`
- [x] `RadialGradient`
- [x] `ImagePaint`
- [x] `SelectionShapeStyle`
- [x] `SeparatorShapeStyle`
- [x] `TintShapeStyle`
- [x] `ForegroundStyle`
- [x] `BackgroundStyle`

Missing modifiers:
- [x] `opacity`
- [x] `blendMode`
- [x] `shadow`


shadowfacts (approved):

Sources/LiveViewNative/Utils/ShapeStyle/ShadowStyle.swift
```diff
@@ -0,0 +1,63 @@
...
+import SwiftUI
+
+/// A shadow configuration.
+///
+/// Two types of shadow can be created, `:drop` and `:inner`. Use a tuple where the first element is the type and the second is a keyword list or map with any options.
+///
+/// The `radius` argument is required.
+///
+/// ```elixir
+/// {:drop, [radius: 10]}
```
supernintendo, line 17: @carson-katri Hmm, I tried this but it doesn't seem to be working:

```heex
<VStack id="hello-ios" modifiers={@native |> foreground_style(primary: {:shadow, {:drop, [radius: 10]}})}>
  <Spacer />
  <HStack modifiers={@native |> padding(all: 5)}>
    <Text>Hello iOS!</Text>
  </HStack>
</VStack>
```

Screenshot 2023-04-25 at 1 21 08 PM

lib/live_view_native_swift_ui/types/shape_style/image_paint.ex
```diff
@@ -0,0 +1,40 @@
...
+            scale: value[:scale] || 1
+          }}
+        {:system, system_image} ->
+          {:ok, %__MODULE__{
+            image: nil,
+            system_image: system_image,
+            source_rect: source_rect,
+            scale: value[:scale] || 1
+          }}
+      end
```
supernintendo, line 32: Just a minor formatting suggestion and returning an `:error` atom as a fallback case so that it can be handled by the modifiers system in `live_view_native`.

```suggestion
      case value[:image] do
        {:name, image} ->
          {:ok, %__MODULE__{
            image: image,
            system_image: nil,
            source_rect: source_rect,
            scale: value[:scale] || 1
          }}

        {:system, system_image} ->
          {:ok, %__MODULE__{
            image: nil,
            system_image: system_image,
            source_rect: source_rect,
            scale: value[:scale] || 1
          }}

        _ ->
          :error        
      end
```

lib/live_view_native_swift_ui/types/shape_style/shadow_style.ex
```diff
@@ -0,0 +1,29 @@
...
+    {:ok, struct(
+      __MODULE__,
+      (if Map.has_key?(with_type, :color),
+        do: Map.replace(with_type, :color, elem(Color.cast(with_type[:color]), 1)),
+        else: with_type
+      )
+    )}
+  end
+  def cast(_), do: :error
+end
```
supernintendo, line 29: This code could be made a bit more readable by breaking the conditional out into a separate function:

```suggestion
  def cast({type, opts}) when type in [:drop, :inner] and is_map(opts) or is_list(opts) do
    with_type =
      opts
      |> Enum.into(%{})
      |> Map.put(:type, type)

    {:ok, struct(__MODULE__, cast_color(with_type))}
  end

  def cast(_), do: :error

  ###

  defp cast_color(%{color: color} = with_type) do
    cast_color = Color.cast(color)

    Map.replace(with_type, :color, elem(cast_color, 1))
  end

  defp cast_color(with_type), do: with_type
```

supernintendo (changes requested):

Sources/LiveViewNative/Utils/ShapeStyle/ShadowStyle.swift
carson-katri: Shadow is a modifier that applies to another shape style, so you would need something like:
```elixir
{:foreground, [{:shadow, {:drop, [radius: 10]}}]}
```

supernintendo (approved):

dginsburg:
@carson-katri Is this ready to be merged?

PR #842, opened 2023-04-21
Fix mismatch between Elixir and Swift values for `text_field_style`

dginsburg:
@carson-katri I was testing other `TextField` modifiers and ran into this mismatch between the Elixir and Swift values. Underscore is what I'm seeing used elsewhere so I changed it to that.


carson-katri (approved):
Good catch, thanks

PR #852, opened 2023-04-24
Add `font_design` modifier

### LINKED ISSUE #300: Modifier -> SwiftUI -> Text input and output: fontDesign(_:)
Doc: https://developer.apple.com/documentation/swiftui/view/fontdesign(_:)

- [ ] Swift implementation
- [ ] Elixir implementation

supernintendo:
Adds the [fontDesign](https://developer.apple.com/documentation/swiftui/view/fontdesign(_:)) text input and output modifier:

```heex
<VStack id="font_design">
  <Text modifiers={font_design(@native, design: :default)}>This text is default</Text>
  <Text modifiers={font_design(@native, design: :monospaced)}>This text is monospaced</Text>
  <Text modifiers={font_design(@native, design: :rounded)}>This text is rounded</Text>
  <Text modifiers={font_design(@native, design: :serif)}>This text is serif</Text>
</VStack>
```

Screenshot 2023-04-24 at 11 30 15 AM

Closes #300

### DIFF
--- a/Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
+++ b/Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
@@ -45,7 +45,11 @@ struct FontDesignModifier: ViewModifier, Decodable {
     }
 
     func body(content: Content) -> some View {
-        content.font(.system(size: UIFont.labelFontSize, design: design))
+        if #available(iOS 16.1, *) {
+            content.fontDesign(design)
+        } else {
+            content
+        }
     }
     
     enum CodingKeys: String, CodingKey {
--- a/Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
+++ b/Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
@@ -45,7 +45,7 @@ struct FontDesignModifier: ViewModifier, Decodable {
     }
 
     func body(content: Content) -> some View {
-        if #available(iOS 16.1, *) {
+        if #available(iOS 16.1, watchOS 9.1, *) {
             content.fontDesign(design)
         } else {
             content
--- a/Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
+++ b/Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
@@ -9,10 +9,12 @@ import SwiftUI
 
 /// Applies a font design to any element.
 ///
+/// ```html
 /// <Text modifiers={font_design(@native, design: :default)}>This text is default</Text>
 /// <Text modifiers={font_design(@native, design: :monospaced)}>This text is monospaced</Text>
 /// <Text modifiers={font_design(@native, design: :rounded)}>This text is rounded</Text>
 /// <Text modifiers={font_design(@native, design: :serif)}>This text is serif</Text>
+/// ```
 ///
 /// ## Arguments
 /// * ``design``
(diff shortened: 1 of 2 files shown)


Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
```diff
@@ -0,0 +1,52 @@
...
+            design = .rounded
+        case "serif":
+            design = .serif
+        default:
+            throw DecodingError.dataCorruptedError(forKey: .design, in: container, debugDescription: "expected valid value for Font.Design");
+        }
+    }
+
+    func body(content: Content) -> some View {
+        content.font(.system(size: UIFont.labelFontSize, design: design))
```
shadowfacts, line 46: This should directly use the `.fontDesign` modifier, since this will override the size if it's set elsewhere and will break dynamic type

shadowfacts (changes requested):

Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
supernintendo: @shadowfacts Fixed. It seems that `.fontDesign` is only available on iOS 16.1 or later so I had to use an `#available` version check. I'm not sure how to handle the fallback in a way that doesn't override the font size if its already set, so I ended up just ignoring it on older versions (feel free to make a suggestion for how to handle that if its something we want to do)

shadowfacts: I agree. It'd be nice if we could mark the the modifier type with `@available` so the docs reflect the actual availability, but the builtin registry generator doesn't support it at the moment (because the availability check would move into the generated code).

shadowfacts (approved):

Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
```diff
@@ -0,0 +1,58 @@
...
+/// <Text modifiers={font_design(@native, design: :monospaced)}>This text is monospaced</Text>
+/// <Text modifiers={font_design(@native, design: :rounded)}>This text is rounded</Text>
+/// <Text modifiers={font_design(@native, design: :serif)}>This text is serif</Text>
+///
+/// ## Arguments
+/// * ``design``
+#if swift(>=5.8)
+@_documentation(visibility: public)
+#endif
+struct FontDesignModifier: ViewModifier, Decodable {
```
carson-katri, line 22: If you put the availability check on the modifier type instead of in the body it should show more prominently in the docs.

```suggestion
@available(iOS 16.1, tvOS 16.1, watchOS 9.1, *)
struct FontDesignModifier: ViewModifier, Decodable {
```

carson-katri (changes requested):

carson-katri (approved):

carson-katri:
I can look into availability checks in the registry generator.

Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+
+import SwiftUI
+
+/// Applies a font design to any element.
+///
+/// <Text modifiers={font_design(@native, design: :default)}>This text is default</Text>
+/// <Text modifiers={font_design(@native, design: :monospaced)}>This text is monospaced</Text>
+/// <Text modifiers={font_design(@native, design: :rounded)}>This text is rounded</Text>
+/// <Text modifiers={font_design(@native, design: :serif)}>This text is serif</Text>
```
carson-katri, line 15: This should be in a codeblock or it won't render.

supernintendo: Oops, good catch

Sources/LiveViewNative/Modifiers/Text Input and Output Modifiers/FontDesignModifier.swift
supernintendo: @shadowfacts I think this is addressed by https://github.com/liveview-native/liveview-client-swiftui/pull/868 and out of scope for this PR anyway so marking this as resolved 🙂

PR #857, opened 2023-04-24
Clean up modifiers directory in the Elixir lib

supernintendo:
This PR cleans up the modifiers directory in the Elixir library by grouping modifiers by their category in the SwiftUI documentation. The directory structure follows what we're doing in the Swift library, except with snake case formatting as is the convention.

Before:
Screenshot 2023-04-24 at 12 37 22 PM

After:
Screenshot 2023-04-24 at 12 35 06 PM


bcardarella:
I love it 😄

Issue #601, opened 2023-03-13
Modifier -> SwiftUI -> Input events: digitalCrownRotation(_:onChange:onIdle:)


AZholtkevych:
Doc: https://developer.apple.com/documentation/swiftui/view/digitalcrownrotation(_:onchange:onidle:)

- [ ] Swift implementation
- [ ] Elixir implementation

AZholtkevych:
We won't support this for now

PR #870, opened 2023-04-26
Add `digital_crown_accessory` modifiers

### LINKED ISSUE #599: Modifier -> SwiftUI -> Input events: digitalCrownAccessory(_:)
Doc: https://developer.apple.com/documentation/swiftui/view/digitalcrownaccessory(_:)

- [ ] Swift implementation
- [ ] Elixir implementation

### LINKED ISSUE #600: Modifier -> SwiftUI -> Input events: digitalCrownAccessory(content:)
Doc: https://developer.apple.com/documentation/swiftui/view/digitalcrownaccessory(content:)

- [ ] Swift implementation
- [ ] Elixir implementation

dginsburg:
Closes #599 
Closes #600 

@carson-katri @supernintendo @shadowfacts
Can you give this PR an extra close look? I don't have a way to test on Apple Watch and I'm not sure I used content tags correctly and put the `digital_crown_accessory_visibility` modifier on the correct element in the doc example I made.

https://developer.apple.com/documentation/swiftui/view/digitalcrownaccessory(_:)
> Use this method to customize the visibility of a Digital Crown accessory View created with the View/digitalCrownAccessory(_ content:) modifier.


Sources/LiveViewNative/Modifiers/Input Events Modifiers/DigitalCrownAccessoryModifier.swift
```diff
@@ -0,0 +1,56 @@
...
+/// Nested content is referenced by its namespace using the `content` argument.
+///
+/// ```html
+/// <List modifiers={@native |> digital_crown_accessory(content: :dca_view)}>
+///     ...
+///     <digital_crown_accessory:dca_view>
+///         <Image system-name="heart.fill" modifiers={@native |> digital_crown_accessory_visibility(visibility: :visible)} />
+///     </digital_crown_accessory:dca_view>
+/// </List>
+/// ```
```
dginsburg, line 21: Not feeling really confident about this example and I'm not sure how to test it.

Sources/LiveViewNative/Modifiers/Input Events Modifiers/DigitalCrownAccessoryModifier.swift
```diff
@@ -0,0 +1,56 @@
...
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.content = try container.decode(String.self, forKey: .content)
+    }
+
+    func body(content: Content) -> some View {
+        content
+            #if os(watchOS)
+            .digitalCrownAccessory {
+                context.buildChildren(of: element, withTagName: self.content, namespace: "digital_crown_accessory")
```
carson-katri, line 48: This modifier resets the environment, so you need to manually propagate the `coordinatorEnvironment` and `anyLiveContextStorage` environment values to avoid crashing when drawing the content.

Sources/LiveViewNative/Modifiers/Input Events Modifiers/DigitalCrownAccessoryModifier.swift
carson-katri: This example mostly looks good, but the `digital_crown_accessory_visibility` modifier needs to go around the `digital_crown_accessory` modifier. The use case for this is fairly narrow, I think it'd be more useful if we supported `digitalCrownRotation`.

carson-katri (dismissed):

Sources/LiveViewNative/Modifiers/Input Events Modifiers/DigitalCrownAccessoryModifier.swift
dginsburg: `digitalCrownRotation` was put in the rejected column a few weeks ago. #601

@carson-katri @AZholtkevych Do we want to add it back?

Sources/LiveViewNative/Modifiers/Input Events Modifiers/DigitalCrownAccessoryVisibilityModifier.swift
```diff
@@ -12,10 +12,13 @@ import SwiftUI
...
-/// <List modifiers={@native |> digital_crown_accessory(content: :dca_view)}>
-///     ...
+/// <List modifiers={
+///     @native
+///     |> digital_crown_accessory(content: :dca_view)
+///     |> digital_crown_accessory_visibility(visibility: :visible)
+/// }>
 ///     <digital_crown_accessory:dca_view>
-///         <Image system-name="heart.fill" modifiers={@native |> digital_crown_accessory_visibility(visibility: :visible)} />
+///         <Image system-name="heart.fill" />
```
dginsburg, line 21: @carson-katri Is this what you meant by `_visibility` needing to be around the `_accessory` view?

Sources/LiveViewNative/Modifiers/Input Events Modifiers/DigitalCrownAccessoryModifier.swift
dginsburg: @carson-katri I copied what looked like similar examples but, tbh, I don't really understand the whole process. Could you help me understand how this is working? If it's too much for text maybe we could screen share and you can walk me through it.

Sources/LiveViewNative/Modifiers/Input Events Modifiers/DigitalCrownAccessoryModifier.swift
```diff
@@ -0,0 +1,59 @@
...
+                    .environment(\.coordinatorEnvironment, coordinatorEnvironment)
+                    .environment(\.anyLiveContextStorage, context)
```
shadowfacts, line 51: This should be `context.storage` since the actual expected type of this environment key is `LiveContextStorage`

Sources/LiveViewNative/Modifiers/Input Events Modifiers/DigitalCrownAccessoryModifier.swift
shadowfacts: The `coordinatorEnvironment` and `anyLiveContextStorage` environment values provide (type-erased) access to the information that's necessary to actually plumb up the live view—primarily a reference to the `LiveViewCoordinator` object, as well as a handful of other things. The various property wrappers, like `@ObservedElement`, rely on getting this information from the environment to function. And so in the handful of places where SwiftUI doesn't propagate the environment, we need to copy these values manually so that the views generated from the child elements have access to them. Right now the two environment keys are bit duplicative and they could probably be merged.

shadowfacts (changes requested):

Sources/LiveViewNative/Modifiers/Input Events Modifiers/DigitalCrownAccessoryModifier.swift
shadowfacts: The `coordinatorEnvironment and `anyLiveContextStorage` environment values provide (type-erased) access to the information that's necessary to actually plumb up the live view—primarily a reference to the `LiveViewCoordinator` object, as well as a handful of other things. The various property wrappers, like `@ObservedElement`, rely on getting this information from the environment to function. And so in the handful of places where SwiftUI doesn't propagate the environment, we need to copy these values manually so that the views generated from the child elements have access to them. Right now the two environment keys are bit duplicative and they could probably be merged.

PR #882, opened 2023-04-28
Synthesizing Decodable

dginsburg:
The goal with this PR is to take advantage of the Swift compiler's ability to synthesize standard `Decodable` implementations. That allows a bunch of boilerplate code to be removed and less code needed in the future.

The next step towards fully embracing synthesizing would be to write `Decodable` extensions for SwiftUI types instead of embedding the switch statement in the init decoder of the parent object. I wrote one example here for `AspectRatioModifier` and `ContentMode`.

There are some places that would be a win because the SwiftUI type is reused. In other situations, the SwiftUI type is only used for one modifier/view and it's nice for code hygiene, but less important.

Slack discussions:
https://dockyard.slack.com/archives/C02E1GA5THB/p1682608237335009
https://dockyard.slack.com/archives/C02E1GA5THB/p1682692104888309

Details:
- Use [JSONDecoder.KeyDecodingStrategy.convertFromSnakeCase](https://developer.apple.com/documentation/foundation/jsondecoder/keydecodingstrategy/convertfromsnakecase) to automatically convert `CodingKeys` from snake to camel case.
- Remove explicit mapping to snake in our custom `CodingKeys` definitions.
- Remove custom `Decodable` and `CodingKeys` implementations where the Swift compiler can synthesize it.


Sources/LiveViewNative/LVN.swift
```diff
@@ -0,0 +1,16 @@
...
+
+import Foundation
+
+struct LVN {
+    static func JSONDecoder() -> JSONDecoder {
+        let decoder = Foundation.JSONDecoder()
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
+        return decoder
+    }
+}
```
dginsburg, line 16: I didn't see an existing place for something like this. I like the syntax because it namespaces the construction: `LVN.JSONDecoder()`. But it can be done in whatever way fits best in this project.

Sources/LiveViewNative/Modifiers/Layout Adjustments Modifiers/FrameModifier.swift
```diff
@@ -59,11 +59,11 @@ enum FrameModifier: ViewModifier, Decodable, Equatable {
...
-        case minHeight = "min_height"
-        case idealHeight = "ideal_height"
-        case maxHeight = "max_height"
+        case minWidth
+        case idealWidth
+        case maxWidth
+        case minHeight
+        case idealHeight
+        case maxHeight
     }
```
dginsburg, line 68: Removing these custom enum strings is what would need to be done across the project at the same time as the switch to using [JSONDecoder.KeyDecodingStrategy.convertFromSnakeCase](https://developer.apple.com/documentation/foundation/jsondecoder/keydecodingstrategy/convertfromsnakecase)

Sources/LiveViewNative/Utils/ContentMode.swift
```diff
@@ -0,0 +1,30 @@
...
+        switch value {
+        case "fill":
+            self = .fill
+        case "fitty":
+            self = .fit
+        default:
+            throw DecodingError.dataCorrupted(.init(codingPath: decoder.codingPath, debugDescription: "Unknown \(String(describing: decoder.codingPath.last)): \(value)"))
+        }
+    }
+}
```
dginsburg, line 30: An example of moving the decoding of a Swift value type out of the modifier and into a type extension.

Sources/LiveViewNative/LVN.swift
carson-katri: This might be better as just a global variable, like:
```swift
let snakeCaseDecoder: JSONDecoder = {
  ...
}()
```
The module provides the namespace, like `LiveViewNative.snakeCaseDecoder` and the instance can be shared instead of creating and configuring it every time its used.

dginsburg: @carson-katri 

The reason I didn't use a single instance of JSONDecoder is that I don't see a guarantee that it's thread safe to possibly have parallel decodings being done by that same instance. The Apple docs say it's "reusable", which isn't necessarily the same as re-entrant. I looked it up and SO discussions do say it's thread safe as implemented today. But that's a Swift library implementation detail and if it's not promised by the docs I chose to be safe

The performance difference is going to be negligible so my suggestion would be to do it the safe way and optimize later when the project is more complete with more tests.

As for the name, I thought that having `JSONDecoder` remain in the name would make it easy for people to understand what it is. You see that and you know what it is; instead of having to go look it up. But I can see why you don't like having a fake namespace when the library already has a real one.

How about a global function that retains the name and requires the real namespace to invoke?

```swift
func JSONDecoder() -> JSONDecoder {
    let decoder = Foundation.JSONDecoder()
    decoder.keyDecodingStrategy = .convertFromSnakeCase
    return decoder
}

LiveViewNative.JSONDecoder().decode(...)
```

In the future, the implementation of `LiveViewNative.JSONDecoder()` could be changed to use a single instance if that proves worthwhile.

Or if you feel strongly about the global var and single instance I'll change it to what you wrote above. You know the full picture better than I do.

dginsburg: That works. 👍🏻 I changed it.

carson-katri (approved):

Sources/LiveViewNative/LVN.swift
shadowfacts: Thread safety isn't important here, modifiers are always decoded on the main thread. Although we're not currently, I think it is worth reusing the JSONDecoder since it's trivially possible.

dginsburg: It's now using a single instance wrapped by `makeJSONDecoder()`.

PR #899, opened 2023-05-02
Remove fatalError wherever possible

carson-katri:
`fatalError`/`preconditionFailure` calls are replaced with throwing methods or `ErrorView` whenever possible. This should prevent the app from requiring a restart whenever something simple goes wrong.

By default, errors are presented with an alert:

In the `CustomRegistry`, implement the `errorView(for:)` method to override this behavior:

```swift
struct AppRegistry: RootRegistry {
    static func errorView(for error: Error) -> some View {
        Label(error.localizedDescription, systemImage: "xmark.diamond.fill")
            .bold()
            .foregroundStyle(.red)
    }
}
```

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -156,7 +156,6 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
         internalState = .notConnected(reconnectAutomatically: false)
     }
     
-    @_spi(NarwinChat)
     public func reconnect() async {
         disconnect()
         await connect()
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -156,6 +156,11 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
         internalState = .notConnected(reconnectAutomatically: false)
     }
     
+    /// Forces the session to disconnect then connect.
+    ///
+    /// All state will be lost when the reload occurs, as an entirely new LiveView is mounted.
+    ///
+    /// This can be used to force the LiveView to reset, for example after an unrecoverable error occurs.
     public func reconnect() async {
         disconnect()
         await connect()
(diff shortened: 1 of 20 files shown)


Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -156,7 +156,6 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
         internalState = .notConnected(reconnectAutomatically: false)
     }
     
-    @_spi(NarwinChat)
     public func reconnect() async {
```
shadowfacts, line 159: Can you add docs for this noting what it does and when it should generally be used?

Sources/LiveViewNative/NavStackEntryView.swift
```diff
@@ -60,7 +60,7 @@ struct NavStackEntryView<R: RootRegistry>: View {
                             self.liveViewModel.cachedNavigationTitle = navigationTitle
                         }
                 } else {
-                    fatalError("State is `.connected`, but no `Document` was found.")
+                    R.loadingView(for: coordinator.url, state: .connectionFailed(LiveConnectionError.missingDocument))
```
shadowfacts, line 63: This is a little weird, since this state isn't actually the coordinator state. If the connection setup code were to check that there's a document, and transition to the connectionFailed/missingDocument state if not, I think force unwrapping in the view body is fine?

Sources/LiveViewNative/Property Wrappers/FormState.swift
```diff
@@ -117,7 +117,7 @@ public struct FormState<Value: FormValue> {
                 return value
             case .form(let formModel):
                 guard let elementName = element.attributeValue(for: "name") else {
-                    fatalError("Expected @FormState in form mode to have element with name")
+                    return initialValue
```
shadowfacts, line 120: Log a warning in this case here and in the setter. This shouldn't be reachable unless the `name` attribute on an element is dynamic (since `resolveMode` validates the presence of the attribute), but in case it happens, it shouldn't fail silently

shadowfacts (changes requested):

Sources/LiveViewNative/NavStackEntryView.swift
carson-katri: I couldn't actually see a path where this state is possible.

shadowfacts (approved):

PR #900, opened 2023-05-02
Add platform metadata to connection params

supernintendo:
This PR proposes including additional metadata through the connection params so that it can be available to the backend LiveView application. Currently this only includes the platform variant (iOS, iPadOS, macOS, watchOS or tvOS) which developers using LiveView Native might want to match on, but it could be expanded to include other information about the client device.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
@@ -307,6 +307,8 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
     private func getDeviceClass() -> String {
         #if os(watchOS)
         return "watch"
+        #elseif os(macOS)
+        return "mac"
         #else
         switch UIDevice.current.userInterfaceIdiom {
         case .phone:
--- a/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
@@ -299,14 +299,36 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
     private func getPlatformMetadata() throws -> String {
         let encoder = JSONEncoder()
         let metadata: [String : String] = [
-            "platform_variant": getPlatformVariant()
+            "device_class": getDeviceClass(),
+            "os_name": getOSName(),
+            "os_version": getOSVersion()
         ]
         let encodedMetadata = try encoder.encode(metadata)
 
         return String(data: encodedMetadata, encoding: .utf8)!
     }
-    
-    private func getPlatformVariant() -> String {
+
+    private func getDeviceClass() -> String {
+        #if os(watchOS)
+        return "watch"
+        #else
+        switch UIDevice.current.userInterfaceIdiom {
+        case .phone:
+            return "phone"
+        case .pad:
+            return "pad"
+        case .mac:
+            return "mac"
+        case .tv:
+            return "tv"
+        default:
+            return "unspecified"
+        }
+        #endif
+    }
+
+
+    private func getOSName() -> String {
         #if os(macOS)
         return "macOS"
         #elseif os(tvOS)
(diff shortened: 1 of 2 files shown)


Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
```diff
@@ -298,6 +300,28 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
...
+    
+    private func getPlatformVariant() -> String {
+        #if os(macOS)
+        return "macOS"
+        #elseif os(tvOS)
+        return "tvOS"
+        #elseif os(watchOS)
+        return "watchOS"
+        #else
+        return UIDevice.current.localizedModel == "iPhone" ? "iOS" : "iPadOS"
```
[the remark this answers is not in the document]
supernintendo: @bcardarella That _is_ something we could pass along to the server, although there might be more to that since what happens when a user rotates the orientation of their device? Do we attempt to listen for that and send an updated `platformMetadata` somehow when it happens?

carson-katri, line 321: The more *idiomatic* way to access this information is with `UIDevice.current.userInterfaceIdiom` 🙃 
```suggestion
        return UIDevice.current.userInterfaceIdiom == .pad ? "iPadOS" : "iOS"
```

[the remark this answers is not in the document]
carson-katri: In general its better to design the UI based on the [`UserInterfaceSizeClass`](https://developer.apple.com/documentation/swiftui/userinterfacesizeclass), rather than a particular interface idiom. For example, when you open an iPad app in split view it will have a `horizontalSizeClass` of `compact`, which is the same as a normal iOS app. In that case, the UI would adjust to behave like on iOS.

supernintendo: I should have noticed :D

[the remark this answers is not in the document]
supernintendo: That makes sense @carson-katri. Do you think my comment above about updating the server's platform metadata on resizes would still apply? I'm thinking about iPadOS which allows you to have two apps open side-by-side and resize them or like newer iPads that have Stage Manager.

Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
```diff
@@ -298,6 +300,28 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
...
     }
     
+    private func getPlatformMetadata() throws -> String {
+        let encoder = JSONEncoder()
+        let metadata: [String : String] = [
+            "platform_variant": getPlatformVariant()
+        ]
+        let encodedMetadata = try encoder.encode(metadata)
+
+        return String(data: encodedMetadata, encoding: .utf8)!
```
shadowfacts, line 310: Is there a reason this needs to be JSON-encoded inside the connectParams? The return type of getPlatformMetadata could just be `Payload` (i.e., `[String: Any]`) and the dictionary could be returned directly

supernintendo: Thanks for the suggestion. I made that change and it simplifies the use of platform metadata from the Elixir side as well. 🙂

Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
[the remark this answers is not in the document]
supernintendo: @carson-katri Agreed. Let's address that as a separate issue for now as it requires more changes than I'd like to introduce with this PR.

Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
```diff
@@ -298,6 +300,54 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
         }
     }
     
+    private func getPlatformMetadata() throws -> Payload {
+        return [
+            "device_class": getDeviceClass(),
```
carson-katri, line 305: Should this be named to match the SwiftUI value? So `user_interface_idiom`?

carson-katri:
It could be helpful to send the version as separate `(major, minor, patch)` integers so it doesn't have to be parsed. You could get that from this property: https://developer.apple.com/documentation/foundation/processinfo/1410906-operatingsystemversion

supernintendo:
Thanks for the review @carson-katri, all of your points should now be addressed 🙂

shadowfacts (approved):

carson-katri (approved):

PR #902, opened 2023-05-02
Add `control_group_style` modifier

carson-katri:
Closes #487

PR #904, opened 2023-05-02
Move safe_area_index.ex to correct folder

dginsburg:
This file got missed when the Elixir modifiers were reorganized.


carson-katri:
I made this change in #902 as well 😄

PR #905, opened 2023-05-03
Update AnyShape to remove snake case keys

dginsburg:
#837 was in flight when I made the JSON decoding changes so the CodingKeys needed to be changed to not specify camel case.

### DIFF
--- a/Sources/LiveViewNative/Utils/ShapeStyle/AngularGradient.swift
+++ b/Sources/LiveViewNative/Utils/ShapeStyle/AngularGradient.swift
@@ -49,7 +49,7 @@ extension AngularGradient: Decodable {
         case gradient
         case center
         case angle
-        case startAngle = "start_angle"
-        case endAngle = "end_angle"
+        case startAngle
+        case endAngle
     }
 }
--- a/Sources/LiveViewNative/Utils/ShapeStyle/AnyShapeStyle.swift
+++ b/Sources/LiveViewNative/Utils/ShapeStyle/AnyShapeStyle.swift
@@ -201,18 +201,16 @@ extension AnyShapeStyle: Decodable {
             self = Self(BackgroundStyle())
         }
         
-        if container.contains(.modifiers) {
-            var modifiers = try container.nestedUnkeyedContainer(forKey: .modifiers)
-            while !modifiers.isAtEnd {
-                let modifier = try modifiers.nestedContainer(keyedBy: CodingKeys.Modifier.self)
-                switch try modifier.decode(Modifier.self, forKey: .type) {
-                case .opacity:
-                    self = Self(self.opacity(try modifier.decode(Double.self, forKey: .properties)))
-                case .blendMode:
-                    self = Self(self.blendMode(try modifier.decode(BlendMode.self, forKey: .properties)))
-                case .shadow:
-                    self = Self(self.shadow(try modifier.decode(ShadowStyle.self, forKey: .properties)))
-                }
+        var modifiers = try container.nestedUnkeyedContainer(forKey: .modifiers)
+        while !modifiers.isAtEnd {
+            let modifier = try modifiers.nestedContainer(keyedBy: CodingKeys.Modifier.self)
+            switch try modifier.decode(Modifier.self, forKey: .type) {
+            case .opacity:
+                self = Self(self.opacity(try modifier.decode(Double.self, forKey: .properties)))
+            case .blendMode:
+                self = Self(self.blendMode(try modifier.decode(BlendMode.self, forKey: .properties)))
+            case .shadow:
+                self = Self(self.shadow(try modifier.decode(ShadowStyle.self, forKey: .properties)))
             }
         }
     }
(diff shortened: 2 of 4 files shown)


Sources/LiveViewNative/Utils/ShapeStyle/AngularGradient.swift
```diff
@@ -49,7 +49,7 @@ extension AngularGradient: Decodable {
         case gradient
         case center
         case angle
-        case startAngle = "start_angle"
-        case endAngle = "end_angle"
+        case startAngle
```
bcardarella, line 52: how does this look in the LVN template?

dginsburg: @bcardarella It's always snake case in the LVN templates. The Swift client is doing an automatic conversion from snake to camel case. That conversion allows us to take advantage of Swift capabilities to generate boilerplate parsing code that we won't have to write anymore.

bcardarella: Ah ok, thank you for the explanation

Sources/LiveViewNative/Utils/ShapeStyle/AnyShapeStyle.swift
```diff
@@ -201,21 +201,22 @@ extension AnyShapeStyle: Decodable {
...
-        while !modifiers.isAtEnd {
-            let modifier = try modifiers.nestedContainer(keyedBy: CodingKeys.Modifier.self)
-            switch try modifier.decode(Modifier.self, forKey: .type) {
-            case .opacity:
-                self = Self(self.opacity(try modifier.decode(Double.self, forKey: .properties)))
-            case .blendMode:
-                self = Self(self.blendMode(try modifier.decode(BlendMode.self, forKey: .properties)))
-            case .shadow:
-                self = Self(self.shadow(try modifier.decode(ShadowStyle.self, forKey: .properties)))
+        if var modifiers = try? container.nestedUnkeyedContainer(forKey: .modifiers) {
```
dginsburg, line 204: There might be a better way to do this, like checking for existence of the key and then letting it parse so errors aren't ignored. This was just the fastest thing I could do to test it out.

supernintendo (approved):
I can confirm this fixes the regression with modifiers that use `AnyShape` that was introduced by https://github.com/liveview-native/liveview-client-swiftui/pull/837:

With `main`:
Screenshot 2023-05-02 at 3 44 02 PM

With `anyshape-keys`:
Screenshot 2023-05-03 at 9 21 54 AM

Sources/LiveViewNative/Utils/ShapeStyle/AnyShapeStyle.swift
```diff
@@ -201,21 +201,23 @@ extension AnyShapeStyle: Decodable {
...
-        while !modifiers.isAtEnd {
-            let modifier = try modifiers.nestedContainer(keyedBy: CodingKeys.Modifier.self)
-            switch try modifier.decode(Modifier.self, forKey: .type) {
-            case .opacity:
-                self = Self(self.opacity(try modifier.decode(Double.self, forKey: .properties)))
-            case .blendMode:
-                self = Self(self.blendMode(try modifier.decode(BlendMode.self, forKey: .properties)))
-            case .shadow:
-                self = Self(self.shadow(try modifier.decode(ShadowStyle.self, forKey: .properties)))
+        if container.contains(.modifiers) {
```
carson-katri, line 204: This check shouldn't be needed, the Elixir side always sets the `modifiers`. Is there some scenario where that fails?

dginsburg: It's not working with the templates in the Scratchpad demo. I gave more context in the PR description up top and there is also a [Slack thread](https://dockyard.slack.com/archives/C02E1GA5THB/p1683067516382179).

carson-katri: I can't reproduce this. Could it have been a version mismatch between the client package and the Elixir package?

supernintendo: @carson-katri I updated the `main` branch of the [scratchboard](https://github.com/liveview-native/scratchboard) repo which should allow you to easily reproduce this. Here's how it currently looks, using the `main` branch of liveview-client-swiftui as a package dependency in the Xcode project:

Screenshot 2023-05-03 at 6 33 37 PM

The changes in this branch seem to fix it (you'll have to rebase `anyshape-keys` off `main` if you're testing, as it doesn't include your template syntax changes from [this PR](https://github.com/liveview-native/liveview-client-swiftui/pull/787) which scratchboard has been updated to use)

dginsburg: I removed the check for the modifiers key. I was having issues with package mismatches in my local environment so maybe it was a me problem.

@supernintendo Does this PR for your as it is now? If so then I'll update the description and we can merge it.

dginsburg: Everything is working fine for me in my environment after May updated the packages. Sorry for the false alarm. I'll merge this PR with only the CodingKey changes.

carson-katri (approved):

PR #906, opened 2023-05-03
Add coverage validation

carson-katri:
This adds a check for 30%+ coverage. We can increase this percentage as we go back and add more tests.


dginsburg (approved):
Now we have to live up to this...

PR #914, opened 2023-05-09
Update tutorial

carson-katri:
This updates the tutorial to fix some errors and add missing steps.


Sources/LiveViewNative/LiveViewNative.docc/Resources/01-01-06-cats_list_live.ex
```diff
@@ -24,6 +24,10 @@ defmodule LvnTutorialWeb.CatsListLive do
     {:ok, assign(socket, cats: @cats)}
   end
 
+  def render(%{platform_id: :web} = assigns) do
+    ~H""
+  end
```
carson-katri, line 29: I did this to avoid adding HTML templates to the tutorial. Does this seem reasonable, or would it be better to include the HTML templates as steps in the tutorial? It may not be that important if we'll be switching to LiveBook for the tutorial anyways.

carson-katri: Ok, I can update it to use the LVN sigil too.

carson-katri:
I've updated the tutorial to use the `~Z` sigil.

shadowfacts (changes requested):
The actual tutorial changes look good, but in the repo base, is there something we can do avoid vendoring all 900-odd icons? It's a lot of extraneous stuff (that I presume comes from the autogenerated phoenix project?) which isn't relevant to us. It'd mean the generated repo differs slightly from what someone following the tutorial creates, but in this instance, I think that's fine.

shadowfacts (approved):

PR #930, opened 2023-05-15
HTML decode the output of `encode_modifiers`

dginsburg:
@carson-katri Thanks for writing the utility. It was a big help.

Since we escape the entire input string in Swift, I thought it would make the tests easier to read and tweak if they weren't HTML encoded.

```
[{&quot;disable&quot;:true,&quot;type&quot;:&quot;autocorrection_disabled&quot;}]
[{&quot;font&quot;:{&quot;modifiers&quot;:[],&quot;properties&quot;:{&quot;design&quot;:&quot;serif&quot;,&quot;style&quot;:&quot;large_title&quot;},&quot;type&quot;:&quot;system&quot;},&quot;type&quot;:&quot;font&quot;}]
```

vs

```
[{"disable":true,"type":"autocorrection_disabled"}]
[{"font":{"modifiers":[],"properties":{"design":"serif","style":"large_title"},"type":"system"},"type":"font"}]
```

I also made the utility strip `///` so I could cut and paste multiline modifier chains from the doc comments.

### DIFF
--- a/util/encode_modifiers.exs
+++ b/util/encode_modifiers.exs
@@ -12,10 +12,10 @@ defmodule ModifierEncoder do
     platform_modifiers: @native.platform_modifiers
 
   defmacro quote_argv() do
-    Code.string_to_quoted!(
-      hd(System.argv())
-      |> String.replace(~r'///', "")
-    )
+    System.argv()
+    |> hd()
+    |> String.replace(~r'///', "")
+    |> Code.string_to_quoted!()
   end
 
   def encode() do
--- a/util/encode_modifiers.exs
+++ b/util/encode_modifiers.exs
@@ -20,9 +20,10 @@ defmodule ModifierEncoder do
 
   def encode() do
     IO.write(
-      Phoenix.HTML.Safe.to_iodata(quote_argv().modifiers)
-      |> IO.iodata_to_binary
-      |> HtmlEntities.decode
+      quote_argv().modifiers
+      |> Phoenix.HTML.Safe.to_iodata()
+      |> IO.iodata_to_binary()
+      |> HtmlEntities.decode()
     )
   end
 end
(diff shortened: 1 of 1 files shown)


util/encode_modifiers.exs
```diff
@@ -11,11 +12,18 @@ defmodule ModifierEncoder do
...
+      |> String.replace(~r'///', "")
+    )
   end
 
   def encode() do
-    IO.write(Phoenix.HTML.Safe.to_iodata(quote_argv().modifiers) |> IO.iodata_to_binary)
+    IO.write(
+      Phoenix.HTML.Safe.to_iodata(quote_argv().modifiers)
+      |> IO.iodata_to_binary
+      |> HtmlEntities.decode
```
supernintendo, line 25: Same here. I also prefer to include the `()` for function calls with 0-arity as its more explicit; we should probably enforce this via `mix format` and Credo checks in the future.

```suggestion
      quote_argv().modifiers
      |> Phoenix.HTML.Safe.to_iodata()
      |> IO.iodata_to_binary()
      |> HtmlEntities.decode()
```

util/encode_modifiers.exs
```diff
@@ -11,11 +12,18 @@ defmodule ModifierEncoder do
     platform_modifiers: @native.platform_modifiers
 
   defmacro quote_argv() do
-    Code.string_to_quoted!(hd(System.argv()))
+    Code.string_to_quoted!(
+      hd(System.argv())
+      |> String.replace(~r'///', "")
+    )
```
supernintendo, line 18: Mixing nested arguments with pipes diminishes the readability of this code. Consider:

```suggestion
    System.argv()
    |> hd()
    |> String.replace(~r'///', "")
    |> Code.string_to_quoted!()
```

supernintendo (changes requested):

supernintendo (approved):

PR #937, opened 2023-05-17
Support `drawingGroup` when necessary for capturing snapshots for visual testing

dginsburg:
The test scaffolding uses [CALayer.render](https://developer.apple.com/documentation/quartzcore/calayer/1410909-render#) to generate the image used for visual comparisons. But that function doesn't render view modifiers such as `contrast` and `brightness`.

The usual function for snapshotting a view is [UIView.drawHierarchy](https://developer.apple.com/documentation/uikit/uiview/1622589-drawhierarchy#), but that requires a host application which isn't a part of this package.

The [drawingGroup](https://developer.apple.com/documentation/swiftui/anyview/drawinggroup(opaque:colormode:)#) modifier can solve this by compositing the view offscreen before rendering. However, it has its own quirks.
>Views backed by native platform views may not render into the image. Instead, they log a warning and display a placeholder image to highlight the error.

This PR adds two new features for view snapshot testing
- `useDrawingGroup` is a param for snapshot matching. It defaults to false.
- Adds the `assertFail` companion functions to `assertMatch`. The need for this is to catch silent regressions due to all the view rendering weirdness detailed above. Without an explicit failing test, we wouldn't know if the tests are correctly detecting failures for these view modifiers that require `drawingGroup`.

### DIFF
--- a/Tests/RenderingTests/Modifiers/NavigationModifiersTests.swift
+++ b/Tests/RenderingTests/Modifiers/NavigationModifiersTests.swift
@@ -15,12 +15,12 @@ final class NavigationModifiersTests: XCTestCase {
         #if os(iOS)
         try assertMatch(
             #"""
-            <TabView modifiers='[{"style":"page_always","type":"tab_view_style"}]'>
-                <Text modifiers='[{"label":"label","type":"tab_item"}]'>
+            <TabView modifiers="[{&quot;style&quot;:&quot;page_always&quot;,&quot;type&quot;:&quot;tab_view_style&quot;}]">
+                <Text modifiers="[{&quot;label&quot;:&quot;label&quot;,&quot;type&quot;:&quot;tab_item&quot;}]">
                     A
                     <Image template="label" system-name="person.crop.circle.fill" />
                 </Text>
-                <Text modifiers='[{"label":"label","type":"tab_item"}]'>
+                <Text modifiers="[{&quot;label&quot;:&quot;label&quot;,&quot;type&quot;:&quot;tab_item&quot;}]">
                     B
                     <Image template="label" system-name="tray.fill" />
                 </Text>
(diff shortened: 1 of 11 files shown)


Tests/RenderingTests/Modifiers/NavigationModifiersTests.swift
```diff
@@ -15,18 +15,18 @@ final class NavigationModifiersTests: XCTestCase {
...
                 </Text>
-                <Text modifiers="[{&quot;label&quot;:&quot;label&quot;,&quot;type&quot;:&quot;tab_item&quot;}]">
+                <Text modifiers='[{"label":"label","type":"tab_item"}]'>
                     B
                     <Image template="label" system-name="tray.fill" />
                 </Text>
             </TabView>
             """#,
-            size: .init(width: 100, height: 100)
+            size: .init(width: 200, height: 200)
```
dginsburg, line 29: This one was weird. The test failed when using `drawingGroup`. The images looked the same to my eyes but the tab items were stacked on top of each other because I presume the container wasn't large enough. Increasing the size allowed it to pass.

Tests/RenderingTests/assertMatch.swift
```diff
@@ -34,6 +35,7 @@ extension XCTestCase {
         environment: @escaping (inout EnvironmentValues) -> () = { _ in },
         size: CGSize? = nil,
         lifetime: XCTAttachment.Lifetime = .deleteOnSuccess,
+        useDrawingGroup: Bool = true,
```
carson-katri, line 38: This should not be the default imo. There are many other Views that are backed by UIKit/AppKit that don't crash. Instead, you end up with an image like this: 
<image src="https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/f6fe7817-2e5f-4ac2-954f-6a4cbf74bf61" width="100" />

All of the List tests are now essentially comparing this image to itself. There's no documentation on if/when Apple will switch a View to/from being backed by a platform view. So we should only enable this if we know it is necessary for a test.

carson-katri (changes requested):

Tests/RenderingTests/assertMatch.swift
dginsburg: Oh, that's not good. I'll switch the default.

dginsburg: @carson-katri Please see the main PR description for how I addressed this.

shadowfacts (approved):

carson-katri (approved):

PR #946, opened 2023-05-19
Add strongly typed JSON values to replace Payload where possible

shadowfacts:
This adds a `JSONValue` type that's limited to JSON primitives, and a number of helpers for expressing JSON values in a more Swift-y way:

```swift
let object: JSONValue = [
	"key1": nil as JSONValue,
	"key2": 3.14,
	"key3": "foo",
	"key4": [
		["bar": false, "baz": 42],
		"qux",
	],
]
```

The goal is to replace `Payload` and `Any`, both for general type-safety and with the intent of providing a better foundation for Core-managed channels, since we'll need the ability to pass JSON across the FFI boundary, which means stronger types (one possibility switching out the `JSONValue` enum with a wrapper of whatever type Core provides, and keeping the `ToJSONValue` protocol to provide better expressibility in Swift).

### DIFF
--- a/Sources/LiveViewNative/BuiltinRegistry+applyBinding.swift
+++ b/Sources/LiveViewNative/BuiltinRegistry+applyBinding.swift
@@ -13,7 +13,7 @@ extension BuiltinRegistry {
     static func applyBinding(
         _ binding: AttributeName,
         event: String,
-        value: Payload,
+        value: JSONValue,
         to view: some View,
         element: ElementNode
     ) -> some View {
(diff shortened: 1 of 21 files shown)


carson-katri:
I renamed `ToJSONValue` to `JSONValueConvertible` which better matches other similar protocols in Swift.

bcardarella:
@carson-katri what is the state of this? Can we close?

bcardarella:
@carson-katri following up on this, it's probably no longer necessary but I don't want to close without confirmation.

carson-katri:
@KronicDeth Will JSON be passed as strings to/from core, or do we need a JSON type? And if we do need a type, would uniffi generate it?

simlay:
> Will JSON be passed as strings to/from core, or do we need a JSON type? And if we do need a type, would uniffi generate it?

Ideally we'll be passing types (rather than strings) to/from Core and those types I think will be generated from uniffi.

bcardarella:
Closing as no longer valid

PR #954, opened 2023-05-24
Fix deeply nested comprehensions

### LINKED ISSUE #55: Nesting for loops crash
Nesting for loops in a heex template deeper than 2 leads to the following crash:

PhoenixLiveViewNative/Fragment.swift:282: Fatal error: Static cannot reference template when templates are not provided
2022-10-01 12:39:50.968780-0700 decoder_bug_ios[20933:2863242] PhoenixLiveViewNative/Fragment.swift:282: Fatal error: Static cannot reference template when templates are not provided

I setup a clean-room app to demonstrate the issue here:  [https://github.com/winescout/lvn_decoder_issue_1](https://github.com/winescout/lvn_decoder_issue_1)

carson-katri:
Closes liveview-native/liveview-native-core#19 
Closes liveview-native/liveview-native-core#18

### DIFF
--- a/Sources/LiveViewNative/Fragment.swift
+++ b/Sources/LiveViewNative/Fragment.swift
@@ -24,10 +24,10 @@ enum Child: Decodable, Equatable {
         }
     }
     
-    fileprivate func buildStringInternal(buffer: inout String, root: Root, templates: Templates?) {
+    fileprivate func buildStringInternal(buffer: inout String, root: Root, templates: Templates?, componentStatics: ComponentStatics?, childIndex: Int?) {
         switch self {
         case .fragment(let frag):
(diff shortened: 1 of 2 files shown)


supernintendo (approved):

bcardarella:
why was this fixed in swift? This is a bug in core and fixing this in swift means we need to fix this bug in each client rather than once in core

carson-katri:
@bcardarella Core does not handle this yet. @simlay is working on bringing it to core.

bcardarella:
@carson-katri when issues are being closed on Core from the swift client that doesn't allow us to track the bug to ensure it is being properly addressed.

We need to hold the line on stop adding things to the Swift client as temporary fixes and deal with the pain of waiting on upstream fixes in Core

PR #963, opened 2023-05-30
Add `draggable` modifier

### LINKED ISSUE #230: Modifier -> SwiftUI -> Drag and drop: draggable(_:)
Doc: https://developer.apple.com/documentation/swiftui/view/draggable(_:)

murtza:
Closes #230

### DIFF
--- a/Sources/LiveViewNative/Modifiers/View Configuration Modifiers/DraggableModifier.swift
+++ b/Sources/LiveViewNative/Modifiers/View Configuration Modifiers/DraggableModifier.swift
@@ -1,0 +1,70 @@
+//
+//  DraggableModifier.swift
+//  
+//
+//  Created by murtza on 30/05/2023.
+//
+
+import SwiftUI
+
+/// Activates this view as the source of a drag and drop operation.
+///
+/// ```html
+/// <Text modifiers={draggable(@native, payload: "ABC", preview: :my_preview)}>
+///     ABC
+///     <Image system-name="heart.fill" template={:my_preview}>
+/// </Text>
+/// ```
+#if swift(>=5.8)
+@_documentation(visibility: public)
+#endif
+struct DraggableModifier<R: RootRegistry>: ViewModifier, Decodable {
+    /// A closure that returns a single instance or a value conforming to Transferable that represents the draggable data from this view.
+    #if swift(>=5.8)
+    @_documentation(visibility: public)
+    #endif
+    private let payload: String
+    
+    /// A View to use as the source for the dragging preview, once the drag operation has begun. The preview is centered over the source view.
+    #if swift(>=5.8)
+    @_documentation(visibility: public)
+    #endif
+    private let preview: String?
+    
+    @ObservedElement private var element
+    @LiveContext<R> private var context
+    @Environment(\.coordinatorEnvironment) private var coordinatorEnvironment
+    
+    init(from decoder: Decoder) throws {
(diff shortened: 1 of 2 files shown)


Sources/LiveViewNative/Modifiers/View Configuration Modifiers/DraggableModifier.swift
```diff
@@ -0,0 +1,70 @@
...
+/// <Text modifiers={draggable(@native, payload: "ABC", preview: :my_preview)}>
+///     ABC
+///     <Image system-name="heart.fill" template={:my_preview}>
+/// </Text>
+/// ```
+#if swift(>=5.8)
+@_documentation(visibility: public)
+#endif
+struct DraggableModifier<R: RootRegistry>: ViewModifier, Decodable {
+    /// A closure that returns a single instance or a value conforming to Transferable that represents the draggable data from this view.
```
carson-katri, line 22: This doc comment doesn't seem quite correct in the context of LiveView Native. They don't need to know about transferable, as only strings are supported here, and the input is not a closure.

Sources/LiveViewNative/Modifiers/View Configuration Modifiers/DraggableModifier.swift
```diff
@@ -0,0 +1,70 @@
...
+            content.draggable(
+                payload,
+                preview:
+                    {
+                    context.buildChildren(of: element, forTemplate: preview)
+                            .environment(\.coordinatorEnvironment, coordinatorEnvironment)
+                            .environment(\.anyLiveContextStorage, context.storage)
+                    }
+            )
+        }else{
```
carson-katri, line 57: formatting nit:
```suggestion
        } else {
```

Sources/LiveViewNative/Modifiers/View Configuration Modifiers/DraggableModifier.swift
```diff
@@ -0,0 +1,70 @@
...
+/// ```html
+/// <Text modifiers={draggable(@native, payload: "ABC", preview: :my_preview)}>
+///     ABC
+///     <Image system-name="heart.fill" template={:my_preview}>
+/// </Text>
+/// ```
+#if swift(>=5.8)
+@_documentation(visibility: public)
+#endif
+struct DraggableModifier<R: RootRegistry>: ViewModifier, Decodable {
```
carson-katri, line 21: Please add `@available` here so the watchOS unavailability is clear in the docs.

carson-katri (changes requested):

Sources/LiveViewNative/Modifiers/View Configuration Modifiers/DraggableModifier.swift
```diff
@@ -45,16 +50,12 @@ struct DraggableModifier<R: RootRegistry>: ViewModifier, Decodable {
     func body(content: Content) -> some View {
         #if !os(watchOS)
```
carson-katri, line 51: if it doesn't support tvOS this could be used:
```suggestion
        #if os(iOS) || os(macOS)
```

carson-katri (approved):

carson-katri:
One other thing you may want to fix, I believe this shouldn't be in the `View Configuration Modifiers` folder, but a new one for `Drag and Drop Modifiers` (according to the categories listed in the title of #230).

PR #964, opened 2023-05-30
File upload support

carson-katri:
This is an attempt at adding support for uploading files with LiveView. Unfortunately, I hit some limitations in `SwiftPhoenixClient` that make this impossible to implement. However, this could provide a good starting point for integrating the Rust channels client in the future.


carson-katri:
Yes, core should allow raw data to be sent like what the official Phoenix JS code supports:
https://github.com/phoenixframework/phoenix/blob/adecbdb17a55763fbbc9d4187cd671f8071a333c/assets/js/phoenix/serializer.js#L31-L53

I tried implementing this, but the Swift package we use tries to convert the data back into a String, which fails.

bcardarella:
I believe we are punting on this PR in favor of file upload support getting into LVN Core. I'll let @simlay and @carson-katri confirm before we close.

PR #965, opened 2023-05-30
Fix frame modifier

carson-katri:
This is a refactor of the `frame` modifier to improve decoding and documentation. Flexible frame decoding is fixed, so you can create elements that fill the parent.


Tests/LiveViewNativeTests/DecodeModifiersTest.swift
```diff
@@ -21,21 +21,6 @@ final class DecodeModifiersTest: XCTestCase {
         XCTAssertEqual(decoded, expected)
     }
     
-    func testDecodeFrame() throws {
```
supernintendo, line 24: Any reason why you're removing this test?

supernintendo (approved):
Nice 🎉

PR #966, opened 2023-05-31
Fix template elements and selection in `List`

carson-katri:
This fixes 3 bugs with `List`:
1. Elements with `template={...}` were rendered as list items and required an `id`. They are now ignored.
2. If no `selection` attribute was present, selection was still possible but not sent to the backend. Now selection is disabled if the attribute is not set to a binding (this applies to `Table` as well)
3. If no `phx-delete`/`phx-move` event was provided, delete/move actions were still enabled. Now they are disabled unless the event attribute is set.

### DIFF
--- a/Sources/LiveViewNative/Property Wrappers/LiveContext.swift
+++ b/Sources/LiveViewNative/Property Wrappers/LiveContext.swift
@@ -133,6 +133,14 @@ public struct LiveContext<R: RootRegistry>: DynamicProperty {
     ) -> [NodeChildrenSequence.Element] {
         element.children().filter(Self.isTemplateElement(template))
     }
+    
+    func children(
+        of element: ElementNode
+    ) -> [NodeChildrenSequence.Element] {
+        element.children().filter {
+            !($0.asElement()?.attributes.contains(where: { $0.name.rawValue == "template" }) ?? false)
+        }
+    }
 }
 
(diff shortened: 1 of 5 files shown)


Sources/LiveViewNative/Property Wrappers/LiveContext.swift
```diff
@@ -133,6 +133,14 @@ public struct LiveContext<R: RootRegistry>: DynamicProperty {
     ) -> [NodeChildrenSequence.Element] {
         element.children().filter(Self.isTemplateElement(template))
     }
+    
+    func children(
+        of element: ElementNode
+    ) -> [NodeChildrenSequence.Element] {
+        element.children().filter {
+            !($0.asElement()?.attributes.contains(where: { $0.name.rawValue == "template" }) ?? false)
```
supernintendo, line 141: A few probably dumb questions:

1. I noticed this function calls `asElement` on each child node before doing the conditional check as opposed to calling `asElement` only on child nodes that contain the `template` attribute; is this necessary and if not, do you think it incurs an additional performance cost that is worth refactoring?
2. Any reason for the ternary instead of using `guard`?

carson-katri: Good call, I simplified the implementation.

supernintendo (approved):

PR #969, opened 2023-05-31
Add `container_shape` modifier

### LINKED ISSUE #517: Modifier -> SwiftUI -> Shapes: containerShape(_:)
Doc: https://developer.apple.com/documentation/swiftui/view/containershape(_:)

- [ ] Swift implementation
- [ ] Elixir implementation

carson-katri:
Closes #517

### DIFF
--- a/Sources/LiveViewNative/Utils/ShapeReference.swift
+++ b/Sources/LiveViewNative/Utils/ShapeReference.swift
@@ -159,8 +159,7 @@ enum ShapeReference: Decodable {
         }
     }
     
-    @_disfavoredOverload
-    func resolve<R: RootRegistry>(on element: ElementNode, in context: LiveContext<R>) -> AnyInsettableShape {
+    func resolveInsettableShape<R: RootRegistry>(on element: ElementNode, in context: LiveContext<R>) -> AnyInsettableShape {
         switch self {
         case let .static(anyShape):
             return AnyInsettableShape(anyShape)
--- a/Sources/LiveViewNative/Utils/ShapeReference.swift
+++ b/Sources/LiveViewNative/Utils/ShapeReference.swift
@@ -62,7 +62,7 @@ import SwiftUI
 @_documentation(visibility: public)
 #endif
 enum ShapeReference: Decodable {
-    case `static`(AnyShape)
+    case `static`(any InsettableShape)
     case key(String)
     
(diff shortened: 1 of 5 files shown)


Sources/LiveViewNative/Utils/ShapeReference.swift
```diff
@@ -153,9 +153,80 @@ enum ShapeReference: Decodable {
...
                     } else {
-                        return AnyShape(shape)
+                        return .init(shape)
                     }
-                }.first ?? AnyShape(Rectangle())
+                }.first ?? .init(Rectangle())
+        }
+    }
+    
+    @_disfavoredOverload
```
supernintendo, line 162: Is this private API stuff? I couldn't find much information about `@_disfavoredOverload` when searching so I'm not sure of how / why its being used here.

carson-katri: There are two methods with `resolve(on:in:)` signatures, one returns `AnyShape`, this one returns `AnyInsettableShape`. In some cases it can be ambiguous as to which one should be used. Adding `@_disfavoredOverload` causes the `AnyShape` method to be preferred unless the call site requires an `InsettableShape` type.

supernintendo: @carson-katri What I'm wondering is if this is a private API that we need to avoid because it will affect apps that use this library and potentially get them rejected from App Store approval? The official Swift docs strongly discourage the use of this attribute ([link](https://github.com/apple/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md#underscored-attributes-reference)).

> There are two methods with resolve(on:in:) signatures, one returns AnyShape, this one returns AnyInsettableShape.

Could we avoid overloading the function name by breaking `.resolve` into separate functions that always return the same type and then calling whichever function we need rather than overloading `.resolve`? I know it would probably require some larger refactoring, but it would eliminate the ambiguity and avoid the need to use `@_disfavoredOverload`.

carson-katri: I've refactored it to avoid using that attribute. It wouldn't affect app approval as its just a compile-time attribute in the Swift language not a private Apple API, but in this case it is possible to avoid it.

supernintendo (approved):

PR #973, opened 2023-06-06
Update for Swift 5.9

carson-katri:
* Fixes a build error with Swift 5.9
* Uses the new `ContentUnavailableView` when connection fails as the default on 2023 platforms.

* Adds a new `#Registries` macro that generates the `Registries` typealias in an `AggregateRegistry`. This is a better alternative to the `Registry2/Registry3/...` types, which require manually creating variants for each number of registries.

<table>

<tr>
<th>Macro</th><th>Expanded</th>
</tr>

<tr>

<td>

```swift
struct AppRegistries: AggregateRegistry {
    #Registries<
        MyRegistry,
        LiveFormsRegistry<Self>,
        AVKitRegistry<Self>,
        MapKitRegistry<Self>
    >
}
```

</td>

<td>

```swift
struct AppRegistries: AggregateRegistry {
    typealias Registries = _MultiRegistry<
        AppRegistry,
        _MultiRegistry<
            LiveFormRegistry<Self>,
            _MultiRegistry<
                PhotoKitRegistry<Self>,
                MapKitRegistry<Self>
            >
        >
    >
}
```

</td>

</tr>

</table>

### DIFF
--- a/Sources/LiveViewNative/LiveView.swift
+++ b/Sources/LiveViewNative/LiveView.swift
@@ -57,11 +57,19 @@ public struct LiveView<R: RootRegistry>: View {
                             ErrorView<R>(html: trace)
                         } else {
                             if #available(iOS 17, macOS 14, tvOS 17, watchOS 10, *) {
+                                #if swift(>=5.9)
                                 SwiftUI.ContentUnavailableView {
                                     SwiftUI.Label("No Connection", systemImage: "network.slash")
                                 } description: {
                                     SwiftUI.Text("The app will reconnect when network connection is regained.")
                                 }
+                                #else
+                                SwiftUI.VStack {
+                                    SwiftUI.Text("Connection Failed")
+                                        .font(.subheadline)
+                                    SwiftUI.Text(error.localizedDescription)
+                                }
+                                #endif
                             } else {
                                 SwiftUI.VStack {
                                     SwiftUI.Text("Connection Failed")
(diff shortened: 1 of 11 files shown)


Sources/LiveViewNative/LiveView.swift
```diff
@@ -56,10 +56,26 @@ public struct LiveView<R: RootRegistry>: View {
...
                             ErrorView<R>(html: trace)
                         } else {
-                            SwiftUI.VStack {
-                                SwiftUI.Text("Connection Failed")
-                                    .font(.subheadline)
-                                SwiftUI.Text(error.localizedDescription)
+                            if #available(iOS 17, macOS 14, tvOS 17, watchOS 10, *) {
+                                #if swift(>=5.9)
+                                SwiftUI.ContentUnavailableView {
+                                    SwiftUI.Label("No Connection", systemImage: "network.slash")
```
supernintendo, line 62: @carson-katri Are these Views able to be configured by the application developer so I can show a custom view when there's no connection, etc?

Sources/LiveViewNativeMacros/RegistriesMacro.swift
```diff
@@ -0,0 +1,82 @@
...
+//
+//  Created by Carson Katri on 6/6/23.
+//
+
+import SwiftCompilerPlugin
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SwiftSyntaxMacros
+
+/// Implementation of the `#Registries` macro, which creates the `Registries` typealias for an `AggregateRegistry`.
```
supernintendo, line 13: 🔥

supernintendo (approved):
Nice!

Sources/LiveViewNative/LiveView.swift
carson-katri: Yes, use the `errorView` method on your `CustomRegistry` to provide a custom interface.

PR #984, opened 2023-06-13
Add `@_spi(LiveViewNative)` for better addon library support

carson-katri:
Some add-on libraries (such as Swift Charts and MapKit) will require access to the types marked here.

In regular user code, these types will still be internal. However, addon libraries can mark imports like so:

```swift
@_spi(LiveViewNative) import LiveViewNative
```

This will give access to any of the types marked in this PR.


carson-katri:
#986 replaced this with a cleaner API.

PR #986, opened 2023-06-15
Add ContentBuilder API

carson-katri:
This adds a new API for supporting DSLs other than SwiftUI in add-on libraries.

Like `View` in SwiftUI, libraries such as Swift Charts and MapKit have protocols like `ChartContent` and `MapContent`.

Supporting building with these elements, along with their own modifiers, can take a lot of boilerplate and could cause things to fall out of sync if changes are not carefully made to each add-on library.

This moves the logic for these into a single protocol, `ContentBuilder` to avoid those problems.
The code to support a DSL is now very simple, and automatically handles the `modifiers` and `template` attributes.

```swift
struct ChartContentBuilder: ContentBuilder {
    typealias Content = any ChartContent
    
    enum TagName: String {
        case barMark = "BarMark"
        case plot = "Plot"
    }
    
    enum ModifierType: String, Decodable {
        case foregroundStyle = "foreground_style"
        case offset
    }
    
    static func lookup<R: RootRegistry>(
        _ tag: TagName,
        element: ElementNode,
        context: Context<R>
    ) -> Content {
        switch tag {
        case .barMark:
            return BarMark(element: element)
        case .plot:
            return Plot<R>(element: element, context: context)
        }
    }
    
    static func empty() -> Content {
        Charts.ChartContentBuilder.buildBlock()
    }
    
    static func reduce(accumulated: Content, next: Content) -> Content {
        Charts.ChartContentBuilder.buildPartialBlock(accumulated: accumulated, next: next)
    }
    
    static func decodeModifier<R: RootRegistry>(
        _ type: ModifierType,
        from decoder: Decoder,
        registry _: R.Type
    ) throws -> any ContentModifier<Self> {
        switch type {
        case .foregroundStyle:
            return try ForegroundStyleModifier(from: decoder)
        case .offset:
            return try OffsetModifier(from: decoder)
        }
    }
}
```

### DIFF
--- a/Sources/LiveViewNative/ContentBuilder.swift
+++ b/Sources/LiveViewNative/ContentBuilder.swift
@@ -438,6 +438,40 @@ public extension ContentBuilder {
         .environment(\.coordinatorEnvironment, context.coordinatorEnvironment)
         .environment(\.anyLiveContextStorage, context.context)
     }
+    /// Builds nested SwiftUI `View` content.
+    ///
+    /// Use this to include SwiftUI `View` elements that are nested within another DSL.
+    ///
+    /// ```swift
+    /// MapContentBuilder.buildView(
+    ///   element.children(),
+    ///   in: context
+    /// )
+    /// ```
+    @ViewBuilder
+    static func buildChildrenViews<R: RootRegistry>(
+        of element: ElementNode,
+        forTemplate template: String? = nil,
+        in context: Context<R>
+    ) -> some View {
+        if let template {
+            ViewTreeBuilder().fromNodes(
+                element.children()
+                    .filter({ $0.attributes.contains(where: { $0.name == "template" && $0.value == template }) }),
+                context: context.context
+            )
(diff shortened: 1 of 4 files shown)


Sources/LiveViewNative/ContentBuilder.swift
```diff
@@ -0,0 +1,533 @@
...
+///     }
+///   }
+/// }
+/// ```
+@propertyWrapper
+public struct ContentBuilderContext<R: RootRegistry>: DynamicProperty {
+    @Environment(\.coordinatorEnvironment) private var coordinatorEnvironment
+    @LiveContext<R> private var context
+    
+    public init() {}
```
supernintendo, line 297: @carson-katri Did you mean to leave this as is?

supernintendo (approved):
Lgtm!

Sources/LiveViewNative/ContentBuilder.swift
carson-katri: Yeah, I need the init to be public but don't need any arguments to it.

PR #987, opened 2023-06-16
Add constructors for various modifiers

supernintendo:
This PR adds support for modifier constructors for various existing modifiers. It requires the changes in https://github.com/liveview-native/live_view_native/pull/21 and https://github.com/liveview-native/live_view_native_platform/pull/7.


carson-katri (approved):
These constructors look good to me 👍 
I'll have to update the VS Code extension to provide autocomplete for these now.

Issue #17, opened 2022-08-12
EEx.function_from_file call missing in store user favourites section.


Mukulch15:
This might cause issues for users who are directly copying the code.

PR #990, opened 2023-06-21
Add `chart_background` modifier

murtza:
Closes [#17](https://github.com/liveview-native/liveview-native-swiftui-charts/issues/17)


carson-katri:
This should be implemented in the Swift Charts add-on library.

PR #992, opened 2023-06-26
Update to use hex.pm for live_view_native and live_view_native_swiftui

dbii:
The tutorial repo pulls from the main branch, and should pull from the hex.pm versions to avoid issues with breaking code changes during LVN development. This changes the tutorial site code to fix that. I am not sure if this fixes the code that builds to https://github.com/liveview-native/ios-tutorial so if there is someplace else to fix that before it is pushed, please let me know and I will update this PR. AFAIK, elixir was already updated to 1.15 last week.


bcardarella:
I'm OK with this. @supernintendo thoughts?

supernintendo (approved):

PR #997, opened 2023-06-29
Add backing storage and caching for `@Attribute`

carson-katri:
This adds caching to the `@Attribute` property wrapper. The value will only be decoded if it has changed.

I've also updated the `modifiers` attribute to be handled with `@Attribute`, so it can get the benefits of the caching. It is especially useful here, as JSON decoding is expensive.


supernintendo (approved):
Nice!

PR #998, opened 2023-06-29
Add public `_ModifierStack` type

carson-katri:
This adds a new type that can be used by add-on libraries that enables modifiers to be passed as an argument to another modifier.

For example, the `chart_plot_style` modifier in Swift Charts accepts modifiers as an argument:

```elixir
chart_plot_style(
  aspect_ratio(1, content_mode: :fit) |> frame(height: 200)
)
```

See https://github.com/liveview-native/liveview-native-swiftui-charts/pull/96 for the implementation of this modifier.


supernintendo (approved):
Looks good to me

PR #1000, opened 2023-06-29
Apply tag modifier from attributes

### LINKED ISSUE #996: `tag` modifier conflicts with `Phoenix.HTML.Tag.tag/1` and `tag/2`
The [tag modifier](https://github.com/liveview-native/liveview-client-swiftui/blob/1cb2df7ba910907741e77adf87cf3cb95c77bfe8/lib/live_view_native_swift_ui/modifiers/view_fundamentals/tag.ex#L4C1-L4C1) conflicts with an [identically named function](https://hexdocs.pm/phoenix_html/Phoenix.HTML.Tag.html#tag/1) that is often inherited by LiveView module via auto-generated `*.Web` modules in Phoenix projects. This issue proposes changing the tag modifier into an attribute supported by all elements (similar to [what we're doing with IDs](https://github.com/liveview-native/liveview-client-swiftui/blob/1cb2df7ba910907741e77adf87cf3cb95c77bfe8/Sources/LiveViewNative/ViewTree.swift#L83)) to avoid this conflict and to reflect how `tag` is typically used.

supernintendo:
Closes #996

### DIFF
--- a/Sources/LiveViewNative/ViewTree.swift
+++ b/Sources/LiveViewNative/ViewTree.swift
@@ -92,9 +92,11 @@ struct ViewTreeBuilder<R: RootRegistry> {
 
     @ViewBuilder
     private func applyTag(element: ElementNode, to view: some View) -> some View {
-        let tag = element.attributeValue(for: "tag")
-
-        view.tag(tag)
+        if let tag = element.attributeValue(for: "tag") {
+            view.tag(Optional<String>.some(tag))
+        } else {
+            view
+        }
     }
     
     @ViewBuilder
--- a/Sources/LiveViewNative/ViewTree.swift
+++ b/Sources/LiveViewNative/ViewTree.swift
@@ -73,22 +73,28 @@ struct ViewTreeBuilder<R: RootRegistry> {
 
         let modified = view.applyModifiers(R.self)
         let bound = applyBindings(to: modified, element: element, context: context)
-        let withIDAndTag = applyIDAndTag(element: element, to: bound)
+        let withID = applyID(element: element, to: bound)
+        let withIDAndTag = applyTag(element: element, to: withID)
 
         return withIDAndTag
             .environment(\.element, element)
             .preference(key: ProvidedBindingsKey.self, value: []) // reset for the next View.
     }
-    
+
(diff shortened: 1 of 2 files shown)


Sources/LiveViewNative/ViewTree.swift
```diff
@@ -73,18 +73,21 @@ struct ViewTreeBuilder<R: RootRegistry> {
...
+        let withIDAndTag = applyIDAndTag(element: element, to: bound)
+
+        return withIDAndTag
             .environment(\.element, element)
             .preference(key: ProvidedBindingsKey.self, value: []) // reset for the next View.
     }
     
     @ViewBuilder
-    private func applyID(element: ElementNode, to view: some View) -> some View {
+    private func applyIDAndTag(element: ElementNode, to view: some View) -> some View {
```
carson-katri, line 84: Can this instead become separate functions `applyID` and `applyTag`?

supernintendo: Sure

Sources/LiveViewNative/ViewTree.swift
```diff
@@ -87,6 +89,13 @@ struct ViewTreeBuilder<R: RootRegistry> {
             view
         }
     }
+
+    @ViewBuilder
+    private func applyTag(element: ElementNode, to view: some View) -> some View {
+        let tag = element.attributeValue(for: "tag")
+
+        view.tag(tag)
```
carson-katri, line 97: Could this modifier be conditionally applied like `id` is? I'm worried that applying it to every element even if it isn't specified could cause some unexpected behaviors with nested elements.

It'd still have to ensure the tag's type is `Optional<String>` though.

```suggestion
        if let tag = element.attributeValue(for: "tag") {
            view.tag(Optional<String>.some(tag))
        } else {
            view
        }
```

carson-katri (approved):

carson-katri:
Can the tag modifier be removed now?

supernintendo:
@carson-katri Yes, do you want me to include that with these changes or in another PR?

carson-katri:
Either way is fine with me

PR #1004, opened 2023-07-06
Add `#LiveView` macro for registering add-on libraries

carson-katri:
This adds a `#LiveView` macro that simplifies `LiveView` creation with add-on registries in Swift 5.9+.

Addons can be provided in an array of metatypes. The aggregate registry will be automatically synthesized, and passed as a generic argument to the `LiveView`.

```swift
struct ContentView: View {
    var body: some View {
        #LiveView(
            .localhost,
            addons: [ChartsRegistry<_>.self, AVKitRegistry<_>.self]
        )
    }
}
```

Previously, an `AggregateRegistry` needed to be created manually:

```swift
struct ContentView: View {
    var body: some View {
        LiveView<MyRegistry>(.localhost)
    }
}

enum MyRegistry: AggregateRegistry {
    #Registries<
        ChartsRegistry<Self>,
        AVKitRegistry<Self>
    >
}
```

This macro automatically expands to the following:

```swift
{ () -> AnyView in
    enum $s7TestBed11ContentViewV4bodyQrvg04LiveD0fMf0_8RegistryfMu_: AggregateRegistry {
        typealias Registries = _MultiRegistry<ChartsRegistry<Self>, AVKitRegistry<Self>>
    }

    return AnyView(LiveView<$s7TestBed11ContentViewV4bodyQrvg04LiveD0fMf0_8RegistryfMu_>(
            .localhost))
}()
```
This allows it to be used within `View.body` like any other `View`.

### DIFF
--- a/Sources/LiveViewNative/LiveView.swift
+++ b/Sources/LiveViewNative/LiveView.swift
@@ -13,10 +13,10 @@ import Combine
 /// Create a ``LiveView`` with a list of addons.
 ///
 /// Use this macro to automatically register any addons.
-/// Specialize the addon registries with ``EmptyRegistry``.
+/// Use a placeholder type (`_`) as the root for each registry.
 ///
 /// ```swift
-/// #LiveView(.localhost, addons: [ChartsRegistry<EmptyRegistry>.self, AVKitRegistry<EmptyRegistry>.self])
+/// #LiveView(.localhost, addons: [ChartsRegistry<_>.self, AVKitRegistry<_>.self])
 /// ```
 ///
 /// - Note: This macro erases the underlying ``LiveView`` to `AnyView`.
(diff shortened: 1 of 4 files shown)


bcardarella:
I like this, one challenge is in the `addons` are we OK with this approach if there are like 30 addons?

carson-katri:
Type checking succeeds when using `_` placeholder types instead of writing out `EmptyRegistry`, so I updated the docs to suggest this.

```swift
struct ContentView: View {
    var body: some View {
        #LiveView(
            .automatic(development: .localhost, production: .custom(URL(string: "myproductionsite.com")!)),
            configuration: .init(navigationMode: .enabled),
            addons: [ChartsRegistry<_>.self, AVKitRegistry<_>.self]
        )
    }
}
```

supernintendo (approved):

PR #1010, opened 2023-07-25
Make `maximumMessageSize` configurable

carson-katri:
This moves the code for `URLSessionTransport` into our library, and makes the `maximumMessageSize` configurable through the `LiveSessionConfiguration`.


bcardarella:
@carson-katri I think this is the PR you were referring to the other day when we hit the message size limit. But considering we'll be deviating from Phoenix Channels Client in the near(ish) future I assume we can close this?

PR #1058, opened 2023-08-08
Add binding name configuration to FormState

carson-katri:
Every element that used `FormState` now has a custom name for its `value-binding` attribute that matches the SwiftUI property's name.

I've also updated the documentation for these elements to reflect the new binding name.

> **Warning**
> This is a breaking change for any apps that used `value-binding`.


Sources/LiveViewNative/Views/Controls and Indicators/Value Inputs/Slider.swift
```diff
@@ -12,14 +12,14 @@ import SwiftUI
...
 /// ```
 ///
 /// Use ``lowerBound`` and ``upperBound`` to specify the range of possible values.
 ///
 /// ```html
 /// <Slider
-///     value-binding="progress"
+///     value="progress"
 ///     lower-bound={-1}
 ///     upper-bound={2}
```
bcardarella, line 24: why can't use use the `in` from the original SwiftUI view and pass a Range?

```
<Slider
  value="progress"
  in={-1..2}
/>
``

carson-katri: This is not possible because `Range` doesn't implement `Phoenix.HTML.Safe` so it can't be encoded as an attribute.

bcardarella: not supported *yet*

supernintendo (approved):

Issue #1060, opened 2023-08-08
`push_navigate` and `live_redirect` causes `SWIFT TASK CONTINUATION MISUSE` error 


supernintendo:
See attached screenshot. This is happening on `main`

![Screenshot 2023-08-08 at 9 39 18 AM](https://github.com/liveview-native/liveview-client-swiftui/assets/5893007/d7437940-89a5-426d-9503-0e02d33d7088)

PR #1062, opened 2023-08-09
Fix invalid condition for disconnect

supernintendo:
I'm getting a `SWIFT TASK CONTINUATION MISUSE` error any time I try to `push_navigate` from a LiveView. This was originally reported in #1060 and fixed by https://github.com/liveview-native/liveview-client-swiftui/pull/1061 but it looks like the following change turned it from an intermittent error to one that is reproducible 100% of the time:

https://github.com/liveview-native/liveview-client-swiftui/pull/1061/files#diff-bae42148b049400c84c0d3c80e4c54378f44137193561bcbed5e8471c0a3ef9eR268

Reverting that line fixes the error which seems completely resolved by the other changes https://github.com/liveview-native/liveview-client-swiftui/pull/1061

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
@@ -265,7 +265,7 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
     
     func disconnect() async {
         if let channel,
-           !channel.isClosed
+           channel.isClosed
         {
             await withCheckedContinuation { continuation in
                 channel.onClose { _ in


Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
```diff
@@ -265,7 +265,7 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
     
     func disconnect() async {
         if let channel,
-           !channel.isClosed
+           channel.isClosed
```
carson-katri, line 268: I'm not sure this is the correct fix. I'll look into this more closely.

PR #1068, opened 2023-08-15
Use events for client-side persistence

carson-katri:
This goes with the PR to `live_view_native` here: https://github.com/liveview-native/live_view_native/pull/33

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LivePersistenceStore.swift
+++ b/Sources/LiveViewNative/Coordinators/LivePersistenceStore.swift
@@ -1,0 +1,44 @@
+//
+//  LivePersistenceStore.swift
+//
+//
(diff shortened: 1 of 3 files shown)


bcardarella:
@carson-katri what is the state of this branch? If this is the Sessionstorage/Localstorage implementation I believe we should be tackling this first from LVN Core.

Core will handle and pass along to the clients store put/fetch requests and it will be up to each client to provide a persistence store that conforms to a given API. We can expand upon this but I think I was speaking to @simlay about the storage feature and we felt it was best to rely on whatever means are bust suited on each native client for the storage rather than a unified storage method in Core. Core will just be the conduit for the message passing between the client and server.

carson-katri:
This can be closed, it can be redone when core gains this ability.

PR #1116, opened 2023-08-22
Handle `live_redirect` outside of events

### LINKED ISSUE #1109: [Bug]: redirect inside handle_event/3 causes "Not Connected" issue on IOS
### What happened?

redirect/2 navigation is not working inside of a handle event.

I've got a simple example that will reproduce the issue. I've defined the "/home" and the "/about" routes and have identical HomeLive and AboutLive liveviews.

```elixir
defmodule Mix4Web.HomeLive do
  use Mix4Web, :live_view
  use LiveViewNative.LiveView

  def mount(_params, _session, socket) do
    {:ok, redirect(socket, to: "/about")}
  end

  def render(%{platform_id: :web} = assigns) do
    ~H"""
    <.button phx-click="home">Home</.button>
    <.button phx-click="about">About</.button>
    """
  end

  def render(%{platform_id: :swiftui} = assigns) do
    ~SWIFTUI"""
    <Text>HOME</Text>
    <Button phx-click="home">Home</Button>
    <Button phx-click="about">About</Button>
    """
  end

  def handle_event("home", _params, socket) do
    {:noreply, redirect(socket, to: "/home")}
  end

  def handle_event("about", _params, socket) do
    {:noreply, redirect(socket, to: "/about")}
  end
end

```

Interestingly, navigation works fine when the redirect occurs in the `mount/3` callback, but not in the `handle_event/3` callback.

### Library Version

main

### XCode Version

14.3.1

### Swift Version

_No response_

iPhone

IOS 16.4

### Relevant log output

_No response_

carson-katri:
Closes #1109 

Previously, `live_redirect` was only handled in response to events. However, when `handle_info` results in a `push_navigate` a `live_redirect` message is sent independent of an event. This adds a handler for this message on all channels.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -389,6 +389,20 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
                     ))
                 }
             }
+        case .patch:
+            // patch is like `replaceTop`, but it does not disconnect.
+            let coordinator = (navigationPath.last?.coordinator ?? rootCoordinator)!
+            coordinator.url = redirect.to
+            let entry = LiveNavigationEntry(url: redirect.to, coordinator: coordinator)
+            switch redirect.kind {
+            case .push:
+                navigationPath.append(entry)
+            case .replace:
+                // If there is nothing to replace, change the root URL.
+                if !navigationPath.isEmpty {
+                    navigationPath[navigationPath.count - 1] = entry
+                }
+            }
         }
     }
 }
--- a/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
@@ -119,6 +119,11 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
         } else if session.configuration.navigationMode.permitsRedirects,
                   let redirect = (replyPayload["live_redirect"] as? Payload).flatMap({ LiveRedirect(from: $0, relativeTo: self.url) }) {
             try await session.redirect(redirect)
+        } else if session.configuration.navigationMode.permitsRedirects,
+                  let redirect = (replyPayload["redirect"] as? Payload),
+                  let destination = (redirect["to"] as? String).flatMap({ URL.init(string: $0, relativeTo: self.url) })
+        {
+            try await session.redirect(.init(kind: .push, to: destination, mode: .replaceTop))
         }
     }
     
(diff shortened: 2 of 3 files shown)


BrooklinJazz:
Would this also handle using `redirect` and `push_patch`?

Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
```diff
@@ -242,6 +242,14 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
                 try! self.handleDiff(payload: message.payload, baseURL: self.url)
             }
         }
+        channel.on("live_redirect") { [weak self] message in
```
BrooklinJazz, line 245: What event does `push_patch` trigger? I assume it's not `live_redirect` but maybe something like `live_patch`?

Same for `redirect`, I'm not sure if this PR will handle redirect or now.

carson-katri: `push_patch` and `redirect` are separate events/replies. I can open separate PRs to handle those.

BrooklinJazz: Awesome. I'm not sure if I should "Approve" since I'm not really a core team member (...yet) but this looks great to me.

Thank you for solving my issue 🙏

bcardarella:
So the behavior we should expect from `redirect` is the same on the web, it would start a new LV render cycle and the end user's experience should be mostly transparent. It's sub-optimal compared to using `push_navigate` and you lose some state related info but as long as the app isn't crashing that's where we want to be.

carson-katri:
Yes, it appears as a normal navigation push to the user. I based the behavior off how the history entries are handled on the web.

* `redirect` - adds a history entry
* `push_patch`/`push_navigate` - adds a history entry unless `replace: true`

When a history entry is added, it appears as a navigation push transition to a new page.
Otherwise, the current page changes to the new one without any navigation transition.

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -389,6 +389,20 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
                     ))
                 }
             }
+        case .patch:
```
BrooklinJazz, line 392: This is a non-blocking comment, just an observation.

It seems like there's a lot of nested conditional logic to the redirect function.
Consider flattening the nested conditions by combining them.

I don't know much about Swift but in Elixir this would be:

```elixir
case int do
 1 -> 
   case letter do
     "a" -> "1a"
     "b" -> "1b"
    end
 2 -> 
    case letter do
     "a" -> "2a"
     "b" -> "2b"
    end
end
```

vs 

```elixir
case {int, letter} ->
  {1,"a"} -> "1a"
  {1,"b"} -> "1b"
  {2,"a"} -> "2a"
  {2,"b"} -> "2b"
end
```

supernintendo (approved):

PR #1122, opened 2023-08-31
Reconnect when returning from background

carson-katri:
In some cases, when left in the background for a few minutes, the app will lose connection. This PR adds a check that reconnects the socket when returning from the background.

It also fixes an issue with channel rejoins where diffs would be improperly applied, leading to a crash.


supernintendo (approved):

bcardarella:
Am I wrong in assuming that the socket reconnect will result in an app state reset? If so then we should explore solutions like this https://fly.io/phoenix-files/restore-liveview-state-on-startup/

carson-katri:
Yes, that’s correct — assigns will be lost. It *might* be possible to keep sending the heartbeat message in a background task, but I’d have to experiment with that a bit more.

PR #1124, opened 2023-09-05
Version 0.1

supernintendo:


carson-katri (approved):
Are we releasing the Swift Package under this version number as well?

bcardarella:
@carson-katri I don't recall where we landed on this wrt the versioning but we can do that for now.

PR #1134, opened 2023-09-19
Add `class` support to custom registry

### LINKED ISSUE #1133: Add support for class names to custom registry
We would like the custom registry to allow clientside handling of modifiers by providing a way to map view modifiers to class names. This will allow add-on libraries to provide custom styles and build an escape hatch from serverside-only modifiers while we work out design and performance issues with the current implementation (see https://github.com/liveview-native/liveview-client-swiftui/issues/1132).

supernintendo:
This PR adds a new `StylesheetProtocol` that can be used to define custom class names within an app or add-on library's registry. These class names allow modifiers to be defined from the Swift client, which should be more performant than sending them from the server. 

To use, define a stylesheet within your app's registry:

```swift
// ContentView.swift
import SwiftUI
import LiveViewNative

struct ContentView: View {
    var body: some View {
        LiveView<MyRegistry>(.localhost)
    }
}

enum MyRegistry: RootRegistry {
    /// An example of a custom registry stylesheet for a LiveView Native app
    struct Stylesheet: StylesheetProtocol {
        let className: String

        init(_ className: String) {
            self.className = className
        }

            switch className {
            /// A static class name
            case "font-bold":
                content.bold()
            /// A flexible class name that uses substring matching
            case let className where className.starts(with: "color-"):
                let index = className.index(className.startIndex, offsetBy: 6)..<className.endIndex
                let colorName = String("system-\(className[index])")

                content.foregroundStyle(Color(fromNamedOrCSSHex: colorName)!)
            /// If class name is undefined, ignore it
            default:
                content
            }
        }
    }
}
```

Then, reference them in your template using `class`:

```elixir
# hello_live.ex
defmodule MyAppWeb.HelloLive do
  use Phoenix.LiveView
  use LiveViewNative.LiveView

  @impl true
  def render(%{platform_id: :swiftui} = assigns) do
    # This UI renders on the iPhone / iPad app
    ~SWIFTUI"""
    <VStack>
      <Text>Plain text</Text>
      <Text class="color-blue">Text with classname</Text>
      <Text class="color-purple font-bold">Text with classnames</Text>
    </VStack>
    """
  end
  
  # ...
end
```

Result:

Screenshot 2023-09-19 at 1 57 38 PM

Closes #1133

### DIFF
--- a/Sources/BuiltinRegistryGenerator/BuiltinRegistryGenerator.swift
+++ b/Sources/BuiltinRegistryGenerator/BuiltinRegistryGenerator.swift
@@ -15,7 +15,7 @@ struct BuiltinRegistryGenerator: ParsableCommand {
     @Argument(transform: { URL(filePath: $0) }) private var output: URL
     @Option(name: .customLong("view")) private var views: [String] = []
     @Option(name: .customLong("modifier")) private var modifiers: [String] = []
-
+    
     static let denylist = [
         "Shape",
         "TextFieldProtocol",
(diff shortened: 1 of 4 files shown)


supernintendo:
@bcardarella Closing this as the stylesheet approach seems more or less validated so I don't think the ability to define client-side class names will be needed.

PR #1137, opened 2023-09-28
Remove `swiftSettings` arguments for documentation

### LINKED ISSUE #1136: [Bug]: Fresh app fails to build due to unsafe build flags
### What happened?

image

### Library Version

0.1.1

### Xcode Version

15.0

### Swift Version

5

iPhone

iOS 17.0

### Relevant log output

_No response_

carson-katri:
Closes #1136 

Some flags were being passed to the compiler to improve our generated documentation output.
However, these flags were causing issues in newer versions of Xcode when using a version number instead of a branch name or commit hash.

The flags have been removed. If we decide to continue using DocC for documentation instead of switching to hexdocs, we may need to find a workaround for this. A few options are:

1. Move the documentation on affected extensions into articles
2. Conditionally add these "unsafe" flags when building documentation
3. See if newer/upcoming versions of Swift emit this documentation by default and use those for building documentation

I think one of these options could be implemented in a separate PR.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,12 +3,6 @@
 
 import PackageDescription
 
-#if swift(>=5.8)
-let swiftSettings = [SwiftSetting.unsafeFlags(["-emit-extension-block-symbols"])]
-#else
-let swiftSettings = [SwiftSetting]()
-#endif
-
 let package = Package(
     name: "LiveViewNative",
     platforms: [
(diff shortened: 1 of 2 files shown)


bcardarella:
@carson-katri does #2 mean that we'd require having to add those unsafe flags for all installs?

carson-katri:
They would only be included in the build when we are running the `build_docs` script. It would exclude them for regular builds of the package.

supernintendo (approved):

PR #1138, opened 2023-10-01
Improve View evaluation and update performance

carson-katri:
# Document Update Handling
On `main`, updates are observed by subscribing each element to a global `elementChanged` publisher. Anytime an update occurs, the elements will [`filter` the result](https://github.com/liveview-native/liveview-client-swiftui/blob/0e0fc6bbe5e95ef308e51551af0889acb09b87b3/Sources/LiveViewNative/Property%20Wrappers/ObservedElement.swift#L126-L137) to see if they should update.

```mermaid
flowchart TD
    C[Coordinator]
    C --> |View 2 changed| P[Element Changed Combine Publisher]
    P --> |2| 1[View 1]
    1 --> |compare id| 1c[ignore]
    P --> |2| 2[View 2]
    2 --> |compare id| 2c[update]
    P --> |2| 3[View 3]
    3 --> |compare id| 3c[ignore]
```

This PR changes the structure so there is [1 publisher per element](https://github.com/liveview-native/liveview-client-swiftui/blob/ba5cc35b091da4a33396e7d1aa6d6644662c43a3/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift#L41). The change event is [only sent to that publisher](https://github.com/liveview-native/liveview-client-swiftui/blob/ba5cc35b091da4a33396e7d1aa6d6644662c43a3/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift#L404), and elements [no longer have to perform any filtering](https://github.com/liveview-native/liveview-client-swiftui/blob/ba5cc35b091da4a33396e7d1aa6d6644662c43a3/Sources/LiveViewNative/Property%20Wrappers/ObservedElement.swift#L121-L123).

```mermaid
flowchart TD
    C[Coordinator]
    C --> P1[Combine Publisher 1]
    C --> |View 2 changed| P2[Combine Publisher 2]
    C --> P3[Combine Publisher 3]
    P1 --> 1[View 1]
    P2 --> |update| 2[View 2]
    P3 --> 3[View 3]
```

# Observation
The [`@ObservedElement`](https://github.com/liveview-native/liveview-client-swiftui/blob/0e0fc6bbe5e95ef308e51551af0889acb09b87b3/Sources/LiveViewNative/Property%20Wrappers/ObservedElement.swift#L53) property wrapper is used to update a View in response to changes.
On `main`, every instance of this property wrapper has an [internal `Observer` type](https://github.com/liveview-native/liveview-client-swiftui/blob/0e0fc6bbe5e95ef308e51551af0889acb09b87b3/Sources/LiveViewNative/Property%20Wrappers/ObservedElement.swift#L113) that [subscribes to the `elementChanged` publisher](https://github.com/liveview-native/liveview-client-swiftui/blob/0e0fc6bbe5e95ef308e51551af0889acb09b87b3/Sources/LiveViewNative/Property%20Wrappers/ObservedElement.swift#L125C1-L125C1). The `@ObservedElement` property wrapper is [used by other property wrappers](https://github.com/liveview-native/liveview-client-swiftui/blob/0e0fc6bbe5e95ef308e51551af0889acb09b87b3/Sources/LiveViewNative/Property%20Wrappers/Attribute.swift#L46), leading to many redundant subscriptions (each with their own `filter` on `main`):

```mermaid
flowchart TD
    View
    View --> O[ObservedElement]
    View --> A1[Attribute]
    A1 --> O1[ObservedElement]
    View --> A2[Attribute]
    A2 --> O2[ObservedElement]

    O --> S[Observer]
    O1 --> S1[Observer]
    O2 --> S2[Observer]
    C[Element Changed Publisher]
    S --> |sink| C
    S1 --> |sink| C
    S2 --> |sink| C
```

In this PR, the structure is simplified so there is [one `Observer` class for the entire `View`](https://github.com/liveview-native/liveview-client-swiftui/blob/ba5cc35b091da4a33396e7d1aa6d6644662c43a3/Sources/LiveViewNative/ViewTree.swift#L54), which [each `@ObservedElement` shares](https://github.com/liveview-native/liveview-client-swiftui/blob/ba5cc35b091da4a33396e7d1aa6d6644662c43a3/Sources/LiveViewNative/Property%20Wrappers/ObservedElement.swift#L56). This one class subscribes to the element changed publisher. This reduces the number of heap allocations and time spent handling updates for each View.

```mermaid
flowchart TD
    View
    View --> S[Observer]
    View --> O[ObservedElement]
    View --> A1[Attribute]
    A1 --> O1[ObservedElement]
    View --> A2[Attribute]
    A2 --> O2[ObservedElement]

    O --> S
    O1 --> S
    O2 --> S
    S --> |sink| C[Element Changed Publisher]
```

# ForEach View count pre-computation
Apple recommends avoiding branching in `ForEach`. For example, instead of this:
```swift
ForEach(items) { item in
  if item != 3 {
    Text("This item is allowed")
  }
}
```
They recommend this:
```swift
ForEach(items.filter({ $0 != 3 })) { item in
  Text("This item is allowed")
}
```
This is because `ForEach` needs to know how many Views it computes to. When you use an `if`, `switch`, or some other conditional rendering method inside `ForEach`, it becomes ambiguous. Each entry in the loop could evaluate to 0, 1, or more Views. This causes it to compute each element in the `ForEach` eagerly. When there is only the possibility of 1 view (like in the second example), SwiftUI is able to compute elements on-demand when nested inside lazy containers (such as `List`, `Lazy*Stack`, `Lazy*Grid`, and `Table`)

[e35d6d8](https://github.com/liveview-native/liveview-client-swiftui/pull/1138/commits/e35d6d8ef10f6a4d885ec6769badf641d8bff9df) offloads the switch into a separate `View` instance. I do not know if this will allow this optimization to fully work, but its the best attempt I can think of at the moment.


bcardarella:
Oh I like this

supernintendo (approved):

PR #1146, opened 2023-10-10
Continue rendering disconnected pages in navigation path

### LINKED ISSUE #1130: Preserve previous document while reconnecting
When the app reconnects (such as during navigation or when returning from the background), the loading state is always displayed. For a more seamless experience, the previous document could stay presented and re-hydrate when the connection is re-established.

One thing to make it more clear that the app is loading could be adding `disabled` around the entire hierarchy during this loading process. This would ensure that the user cannot interact with the app while the connection is dead.

carson-katri:
Closes #1130

- ~~Provide the same functionality for `push_patch`~~ - `push_patch` is functional, but will instead show the current page due to the shared channel.
- [x] Navigating backward by multiple pages is broken (long press on back button)
- [x] Test `redirect` support

This persists the `Document` for disconnected pages, and ensures there is one `LiveViewCoordinator` per navigation history entry.

Previously, navigating backward would temporarily show a blank page:

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/08d829cd-16d2-4719-a251-970b7fbcb043

With these changes, there is now a disabled version of the page still visible, which is then replaced with the newly rendered page after it reconnects:

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/e4a25717-3692-4b54-a82d-b09527ca0c65

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -389,7 +389,8 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
             }
         case .patch:
             // patch is like `replaceTop`, but it does not disconnect.
-            let coordinator = LiveViewCoordinator(session: self, url: redirect.to, from: navigationPath.last?.coordinator ?? rootCoordinator)
+            let coordinator = (navigationPath.last?.coordinator ?? rootCoordinator)!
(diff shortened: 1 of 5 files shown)


simlay (approved):
I spent ~30 minutes trying to find something wrong with this but could not.

It's not in the scope of this issue but I feel like it'd be useful to be able to display a banner someplace saying "Reconnecting" when `next.last!.coordinator.state` `.notConnected`.

PR #1148, opened 2023-10-11
Support tvOS targets

carson-katri:
Currently tvOS simulators are not supported due to limitations in Rust.


Sources/LiveViewNative/Views/Controls and Indicators/Value Inputs/Stepper.swift
```diff
@@ -66,6 +67,7 @@ struct Stepper<R: RootRegistry>: View {
     @Attribute("upper-bound") private var upperBound: Double?
     
     public var body: some View {
+        #if !os(tvOS)
```
simlay, line 70: Hmm, if you have `@available(iOS 13.0, macOS 10.15, watchOS 9.0, *)` above on the struct `Stepper` do you need this `#fi !os(tvOS)`?

simlay (approved):
I think this looks good. It's been my experience with conditional compilation stuff that it's good to add a CI build rule.

The existence of `Package.swift` and `Package@swift-5.9.swift` isn't my favorite but I think there's another PR for that.

Hmm, when I run:
`xcodebuild test -scheme LiveViewNative-Package -sdk appletvsimulator17.0 -destination "OS=17.0,name=Apple TV,arch=x86_64" -enableCodeCoverage YES -resultBundlePath TestResults`
I get
```
Testing failed:
        'Slider' is unavailable in tvOS
        'init(value:in:onEditingChanged:)' is unavailable in tvOS
        'Slider' is unavailable in tvOS
        'init(value:in:label:minimumValueLabel:maximumValueLabel:onEditingChanged:)' is unavailable in tvOS
        'Stepper' is unavailable in tvOS
        'init(value:step:onEditingChanged:label:)' is unavailable in tvOS
        Testing cancelled because the build failed.
```

I spent a little bit of time trying to add `@available(...)` to `ValueInputTests` and it's unclear if [the `@available` attribuet works with `XCTestCase`s](https://stackoverflow.com/a/63723825)

Also, I'm not sure the `arch=x86_64` attribute was going to build the tests for x86 anyway. Having conditional compilation in CI is a big plus to me. I'd suggest adding it now as the github runners aren't M1's (yet) but then when (not if) they break, you won't have a way to run the tests on your M1.

Sources/LiveViewNative/Views/Controls and Indicators/Value Inputs/Stepper.swift
carson-katri: The `#if` excludes it at compile time to prevent build errors.

carson-katri:
@simlay This repo uses a MacStadium self-hosted runner, so I don't think I can run the test suite on tvOS. I added a "Build for tvOS" step though, which should at least make sure we don't break compilation.

simlay:
> @simlay This repo uses a MacStadium self-hosted runner, so I don't think I can run the test suite on tvOS. I added a "Build for tvOS" step though, which should at least make sure we don't break compilation.

LGTM.

PR #1149, opened 2023-10-12
Implement rules parser

NduatiK:


mix.exs
```diff
@@ -26,7 +26,8 @@ defmodule LiveViewNative.SwiftUI.MixProject do
       {:ex_doc, ">= 0.0.0", only: :dev, runtime: false},
       {:jason, "~> 1.2"},
       {:live_view_native_platform, "~> 0.1"},
-      {:live_view_native_stylesheet, github: "liveview-native/live_view_native_stylesheet"},
+      {:live_view_native_stylesheet, github: "NduatiK/live_view_native_stylesheet", branch: "nk-refactor-RulesParser.Helpers"},
```
bcardarella: now that https://github.com/liveview-native/live_view_native_stylesheet/pull/12 is merged let's revert this LOC

test/live_view_native/swiftui/rules_parser_test.exs
```diff
@@ -259,9 +259,9 @@ defmodule LiveViewNative.SwiftUI.RulesParserTest do
     end
 
     test "to_ime in the middle of a chain" do
-      input = "color(foo.to_ime(color).bar)"
+      input = "color(.foo.to_ime(color).bar)"
```
bcardarella, line 262: this was correct as-is. Some imes can be chained from somethign else. Technically, you'd see something like `color(color: Color.red)`

lib/live_view_native/swiftui/rules_parser.ex
```diff
@@ -1,5 +1,6 @@
 defmodule LiveViewNative.SwiftUI.RulesParser do
   use LiveViewNative.Stylesheet.RulesParser, :mock
```
bcardarella, line 2: hmmm... I think this is my fault I don't know why I put this in there. I'll fix after this PR is merged

test/live_view_native/swiftui/rules_parser_test.exs
NduatiK: I am handling the special case of module/class names ([here](https://github.com/liveview-native/liveview-client-swiftui/pull/1149/files#diff-d2bfbf65b14e1f56662259c0c43e3ebd3d0674ccaa279ef368a62c4d4f94b42fR32)). Right now, `color(foo.to_ime(color).bar)` is incorrect but `color(Foo.to_ime(color).bar)` is okay.

Would you want to support both?

bcardarella: @NduatiK is the issue in this case that `foo` is lower-case? If so I can check with @carson-katri to see if this format is valid or not

mix.exs
NduatiK: Done

test/live_view_native/swiftui/rules_parser_test.exs
bcardarella: @NduatiK let's change this to `Foo.to_ime(color).bar`

NduatiK: We can support both, but I expect that we will only have the class name format. It would be strange to see `color(red.to_ime(shade))`. Eventually, we will want to become strict about what is supported so that syntax validation is easier.

bcardarella: agreed, you can change this test to the `Foo` format in the PR

PR #1153, opened 2023-10-19
Change connection format

### LINKED ISSUE #1131: Change connection format
When the client connects it sends over:

```elixir
def render(
        %{native: %{platform_config: %{user_interface_idiom: ui_idiom}}, platform_id: :swiftui} =
          assigns)
          
```

Instead the client should move over to this form:

```elixir
def render(%{platform: platform, target: target} = assigns) do

```

now we can simply do things like:

```elixir
def render(%{platform: :swiftui, target: :watchos} = assigns) do

```

```elixir
def render(%{platform: :swiftui, target: target} = assigns) when target in ~w{ipad macos}a do

```

supernintendo:
This PR updates the initial connection parameters to conform to the changes in https://github.com/liveview-native/live_view_native/pull/60

Closes #1131


Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -409,14 +409,36 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
...
         [
-            "os_name": getOSName(),
+            "app_version": getAppVersion(),
+            "app_build": getAppBuild(),
+            "bundle_id": getBundleID(),
+            "format": "swiftui",
+            "os": getOSName(),
             "os_version": getOSVersion(),
-            "user_interface_idiom": getUserInterfaceIdiom()
+            "target": getTarget()
```
carson-katri, line 420: [Target has another meaning in Xcode](https://developer.apple.com/documentation/xcode/configuring-a-new-target-in-your-project), which could lead to some confusion if we ask people to "check your target for xyz" in documentation/support.

Just providing the `os` would be sufficient in most cases. We could add a `horizontalSizeClass` parameter as well to allow changes based on iPhone/iPad size.

bcardarella: I don't think we've had a formal discussion on this yet. We'll likely run into name collision between different platforms and how they use terminology. WIthin each client we should probably try to adhere to whatever the accepted nomenclature that is used in that ecosystem. On the Elixir side is where we'll have to define a single language on what means what. I'm open to discussing a different term than `target` on the Elixir side to differentiate between the different devices within a platform family but we should figure that out quickly, or just stick with `target`. As @carson-katri you pointed out we can derive that information in Elixir from the OS but we'll need to ensure we can do that for each platform.

carson-katri: Sticking with `target` is fine, I just wanted to bring up the collision.

bcardarella: @carson-katri yup, it's good to point this out. Like I said we can probably drop `target` from the payload and perhaps derive this from the OS value in the response on the Elixir side and inject `target` for matching as part of the response object. Thoughts?

/cc @supernintendo

carson-katri: The current implementation seems better, I meant the user would just use `os` instead of `target`. The Elixir side would not be able to infer iPad/iPhone.

carson-katri (approved):

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
supernintendo: `os` isn't suitable for matching on since both iPhone and iPad report `"iOS"` as the host OS. I also don't consider this to be a meaningful collision as its only limited to the client-to-server integration path and doesn't affect Xcode targets. `target` can have a different meaning across contexts (in the Swift client we have `LiveSessionState` and `LiveSessionCoordinator` which has nothing to do with [live sessions in Phoenix](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.Router.html#live_session/3)).

A few other thoughts:

1. The most descriptive and accurate name would be something like `ui_type` since it is used more to describe the form factor of the target UI rather than the underlying OS. On Android, we might end up with targets like `half_opened` or `full` to provide support for [foldable devices](https://developer.android.com/guide/topics/large-screens/learn-about-foldables) which doesn't really make sense when represented as `os`.

2. It's important to remember that this new set of connection parameters is a cross-platform spec and no longer opinionated toward SwiftUI like before. We could align with Apple's naming conventions but what if we end up conflicting with those on Android or Windows or other platforms? I agree with @bcardarella in that we won't be able to conform perfectly to every native platform's naming conventions and we should allow ourselves some freedom to define our own APIs.

supernintendo:
Merging this and https://github.com/liveview-native/live_view_native/pull/60. We still have time to tweak some naming before 0.2 if we want.

PR #1157, opened 2023-10-26
Add error reporting for stylesheet syntax

NduatiK:


lib/live_view_native/swiftui/rules_parser/tokens.ex
```diff
@@ -123,101 +105,67 @@ defmodule LiveViewNative.SwiftUI.RulesParser.Tokens do
...
-        parsec(:attr),
-        parsec(:event),
-        parsec(:ime),
-        parsec(:helper_function),
-        parsec(:nested_attribute),
-        parsec(:key_value_list),
-        variable()
-      ])
+  def module_name() do
+    expected(
```
bcardarella, line 157: I suspect this combinator is being used for `Color.red` referencing `Color` as a "module", if so that isn't going to be the correct term on the Swift side. In this particular case `Color` is a type and I suspect in nearly all cases we'd be referencing types but I'd want @carson-katri to confirm/deny

NduatiK: I've renamed it to `type_name`

bcardarella (approved):

PR #1159, opened 2023-10-27
Parser should ignore comments

bcardarella:
Once https://github.com/liveview-native/live_view_native_stylesheet/pull/18 is resolved this PR needs to update the dep for the stylesheet lib then satisfy the rules parser error by ignoring comment lines


NduatiK:
Addressed in #1161

PR #1169, opened 2023-11-15
Update to use `self-hosted` tag for CI

carson-katri:
Coverage dropped a bit, I lowered the threshold for now and more tests can be added for stylesheet parsing in separate PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         shell: bash
         run: |
           xcodebuild -scheme BuiltinRegistryGenerator -destination "platform=macOS"
-          xcodebuild test -scheme LiveViewNative-Package -destination "OS=17.0,name=iPhone 15 Pro" -enableCodeCoverage YES -resultBundlePath TestResults -skipMacroValidation -skipPackagePluginValidation
+          xcodebuild test -scheme LiveViewNative-Package -destination "platform=iOS Simulator,OS=17.0,name=iPhone 15 Pro" -enableCodeCoverage YES -resultBundlePath TestResults -skipMacroValidation -skipPackagePluginValidation
       - uses: kishikawakatsumi/xcresulttool@v1
         with:
           path: TestResults.xcresult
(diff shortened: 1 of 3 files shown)


Tests/LiveViewNativeStylesheetMacrosTests/MacroTests.swift
```diff
@@ -27,12 +27,97 @@ final class MacroTests: XCTestCase {
...
+                                    "]" .utf8
+                                }
+                            }
+                            Always((Double?.none))
+                            }
+                            "]" .utf8
+                        }
+                           .parse(&input)} catch {
+                           input = copy}
+            #endif
```
simlay, line 113: Nit: There's some weird indentation in some of this. Unclear if that's desired.

carson-katri: The macro messes up the formatting a bit. I can try to clean up the formatting in the generated code, but I think that'd belong in a separate PR.

.github/workflows/ci.yml
```diff
@@ -6,23 +6,22 @@ on:
...
+    runs-on: self-hosted
     steps:
       - uses: actions/checkout@v2
       - name: Run tests on iOS Simulator
         shell: bash
         run: |
-          sudo xcode-select --switch /Applications/Xcode_14.3.app
           xcodebuild -scheme BuiltinRegistryGenerator -destination "platform=macOS"
-          xcodebuild test -scheme LiveViewNative-Package -sdk iphonesimulator16.4 -destination "OS=16.4,name=iPhone 14 Pro" -enableCodeCoverage YES -resultBundlePath TestResults
+          xcodebuild test -scheme LiveViewNative-Package -destination "platform=iOS Simulator,OS=17.0.1,name=iPhone 15 Pro" -enableCodeCoverage YES -resultBundlePath TestResults -skipMacroValidation -skipPackagePluginValidation
```
simlay, line 24: It's unclear if you need the `OS=17.0.1` here. In the [uniffi branch I'm working on in liveview-native-core, I've found it wasn't needed](https://github.com/liveview-native/liveview-native-core/blob/1e7849540783cf8a9ae749aabe165c924feed853/Makefile.toml#L377-L386). If you don't specify the OS version, it'll use the default simulator runtime/sdk that's installed. So, when it's upgraded, this CI won't need to be updated.

It was also my experience that it doesn't exit with a non-zero status field when it was slightly wrong. :/

simlay (approved):
LGTM. My comments aren't blocking just some thoughts.

PR #1172, opened 2023-11-29
Define default app and root layouts

supernintendo:
This PR defines default `:app` and `:root` layouts so that LVN apps don't break entirely when these aren't defined. Related PR: https://github.com/liveview-native/live_view_native/pull/68


supernintendo:
@carson-katri I think [`layout-navigation-support`](https://github.com/liveview-native/liveview-client-swiftui/tree/layout-navigation-support) needs to be rebased off of `main`, lmk when you do that and I can rebase this branch as well.

carson-katri:
@supernintendo I rebased `layout-navigation-support`.

carson-katri (approved):

PR #1176, opened 2023-12-06
Add template project generation task

### LINKED ISSUE #179: Custom Xcode Project Template for LVN
We should provide a template for creating new LiveView Native projects in XCode.

Even though mixing LiveView into a SwiftUI app is easy providing the project template will give us an opportunity to control how projects should be organized and make it one step easier.

Exposing the registry should be part of the template as well.

supernintendo:
This PR adds a new `xcodegen` based Mix task for generating Xcode projects when running `mix lvn.install`. This is an improvement over the static template project approach we're using in 0.1. Depends on https://github.com/liveview-native/live_view_native/pull/75

Closes #179


priv/xcodegen/Sources/TemplateApp/ContentView.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+
+import SwiftUI
+import LiveViewNative
+
+struct ContentView: View {
+    var body: some View {
+        LiveView(
+            .automatic(development: .localhost, production: .custom(URL(string: "example.com")!)),
+            configuration: .init(connectParams: { _ in ["apns_token": "tokenValue"] })
```
carson-katri, line 13: Not sure if `configuration` should be included by default, this line doesn't send any meaningful data.

lib/mix/tasks/install.ex
```diff
@@ -0,0 +1,175 @@
...
+      System.find_executable("brew") ->
+        status_message("running", "brew install xcodegen")
+        System.cmd("brew", ["install", "xcodegen"])
+        true
+
+      # Install with Mint
+      System.find_executable("mint") ->
+        status_message("running", "mint install yonaskolb/xcodegen")
+        System.cmd("mint", ["install", "yonaskolb/xcodegen"])
+        true
```
carson-katri, line 67: I think it should check for `mint` before `brew`. It's more niche, so if someone has it installed they would probably prefer it for Swift tools like this.
Or, provide an option for how to install it. Some users may prefer to have it contained in the `_build` directory.

lib/mix/tasks/install.ex
```diff
@@ -0,0 +1,175 @@
...
+
+  defp copy_xcodegen_files(%{native_path: native_path}) do
+    priv_dir = :code.priv_dir(:live_view_native_swift_ui)
+    native_project_dir = Path.join(native_path, "swiftui")
+    xcodegen_path = Path.join(priv_dir, "xcodegen")
+
+    File.cp_r(xcodegen_path, native_project_dir)
+  end
+
+  defp replace_app_name_in_sources(%{app_namespace: app_namespace, native_path: native_path}) do
```
carson-katri, line 92: Can something similar be done for the `example.com` URL?

lib/mix/tasks/install.ex
```diff
@@ -0,0 +1,175 @@
...
+        type: :confirm,
+        label: "Xcodegen is required to generate an Xcode project for your app. Would you like to install it?",
+        ignore: System.find_executable("xcodegen") != nil,
+        on_yes: &install_xcodegen/0,
+        on_no: &skip_swiftui_install/0
+      },
+      targets: %{
+        type: :multiselect,
+        label: "Select any number of deployment targets for your app.",
+        options: ["iOS", "macOS", "watchOS"],
```
carson-katri, line 46: Can a tvOS option be added here?

lib/mix/tasks/install.ex
supernintendo: Oops forgot about this. Most definitely.

lib/mix/tasks/install.ex
carson-katri: I don't think we'll be updating core until v0.3. I'm fine with leaving tvOS out from auto-generation until its fully supported.

carson-katri (approved):

PR #1179, opened 2023-12-06
Improve stylesheet parsing failure

carson-katri:
Previously, when a class failed to parse, the entire stylesheet would be ignored.

This PR stops early when it fails to parse a class, but still returns the classes that succeeded.

For example, this stylesheet has an invalid `font` modifier:
```elixir
"bold" do
  bold(true)
end

"font" do
  font(.nonExistentFont)
end
```

```html
<Text class="bold font">Hello, world!</Text>
```

This results in the following message:

```
Stylesheet parsing failed for class `font`:

error: multiple failures occurred

error: /Users/carson.katri/Documents/LiveViewNative/ElixirConf23Apps/elixirconf_chess/lib/elixirconf_chess_web/styles/app_styles.ex:ElixirconfChessWeb.Styles.AppStyles:10: No matching clause found for modifier 'font'. Expected one of `font(_:)`
 --> input:1:433
1 | …Styles.AppStyles], [{:., [file: "/Users/carson.katri/Documents/LiveViewNat…
  |                      ^

error: unexpected input
 --> input:1:233
1 | …true]}], "font" => [{:font, [file: "/Users/carson.katri/Documents/LiveView…
  |                      ^ expected "]"

in stylesheet:

%{"bold" => [{:bold, [file: "/Users/carson.katri/Documents/LiveViewNative/ElixirConf23Apps/elixirconf_chess/lib/elixirconf_chess_web/styles/app_styles.ex", line: 6, module: ElixirconfChessWeb.Styles.AppStyles], [true]}], "font" => [{:font, [file: "/Users/carson.katri/Documents/LiveViewNative/ElixirConf23Apps/elixirconf_chess/lib/elixirconf_chess_web/styles/app_styles.ex", line: 10, module: ElixirconfChessWeb.Styles.AppStyles], [{:., [file: "/Users/carson.katri/Documents/LiveViewNative/ElixirConf23Apps/elixirconf_chess/lib/elixirconf_chess_web/styles/app_styles.ex", line: 10, module: ElixirconfChessWeb.Styles.AppStyles], [nil, :nonExistentFont]}]}]}
```

But the text is still displayed in `bold`, as that class finished parsing before the `font` class failed:

### DIFF
--- a/Sources/LiveViewNative/Stylesheets/StylesheetParser.swift
+++ b/Sources/LiveViewNative/Stylesheets/StylesheetParser.swift
@@ -1,27 +1,55 @@
 import SwiftUI
 import LiveViewNativeStylesheet
 import Parsing
+import OSLog
+
+private let logger = Logger(subsystem: "LiveViewNative", category: "Stylesheet")
 
 struct StylesheetParser<M: ViewModifier & ParseableModifierValue>: Parser {
     let context: ParseableModifierContext
-    
-    var body: some Parser<Substring.UTF8View, [String:[M]]> {
-        "%{".utf8
-        Many(into: [String:[M]]()) { sheet, pair in
-            let (name, value) = pair
-            sheet[name] = value
-        } element: {
-            Whitespace()
-            String.parser(in: context)
-            Whitespace()
-            "=>".utf8
-            Whitespace()
-            Array<M>.parser(in: context)
-            Whitespace()
-        } separator: {
-            ",".utf8
-        } terminator: {
-            "}".utf8
+
+    func parse(_ input: inout Substring.UTF8View) throws -> Dictionary<String, Array<M>> {
+        let fullStylesheet = input
+        try "%{".utf8.parse(&input)
+        // early exit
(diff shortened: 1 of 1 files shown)


bcardarella:
@carson-katri  the stylesheet should be including annotation data in the middle element of the tuple. This will include the original file path and LOC as well as the original syntax. Can you use this to produce the logged error in addition to the AST node?

carson-katri:
The error message does include the file and line number. But the original syntax isn’t sent, so I can’t display that.

bcardarella:
@carson-katri I'm looking at a single error or two errors? If it is two errors it says `error: unexpected input`

NduatiK:
No, we didn't. It's not a difficult addition, though, if we decide to add it.

carson-katri:
@bcardarella It's only one error with multiple failed parsing paths. They both happened on the same line, which is shown at the top. It has to short-circuit when it encounters malformed data, so it can only really show the first error it encounters.

bcardarella:
@NduatiK let's add that then. Thanks!

@carson-katri ok, thanks

PR #1189, opened 2023-12-12
Remove `Array<Self>` from parsers

carson-katri:
Unfortunately, I still have no idea *why* this crashes the compiler (and only in release mode), but removing `Array<Self>` from the parsers seems to fix it.

I have also updated the `_GeneratedModifiers.swift` for the latest Xcode 15.1, changed `xrOS` to `visionOS` across the project, and replaced `Package.swift` with `Package@swift-5.9.swift` as 5.9 is the only supported Swift version.

Swift bug report: https://github.com/apple/swift/issues/70410


Sources/LiveViewNative/Stylesheets/Modifiers/Overrides/SearchScopesModifier.swift
```diff
@@ -22,7 +22,7 @@ struct _SearchScopesModifier<R: RootRegistry>: ViewModifier {
     @ObservedElement private var element
     @LiveContext<R> private var context
     
-    #if os(iOS) || os(macOS) || os(tvOS) || os(xrOS)
+    #if os(iOS) || os(macOS) || os(tvOS) || os(visionOS)
```
simlay, line 25: This is a more reasonable OS name.

Package@swift-5.9.swift
simlay, line 1: Cool.

simlay (approved):
I understand and approve. ~of everything but `Array<Self>.parse`.~ The part where `Array<Self>.parse` fixes a compiler build confuses me.

Sources/LiveViewNative/Stylesheets/Modifiers/Overrides/SearchScopesModifier.swift
carson-katri: That was true of earlier betas, but Xcode now warns about `xrOS` not being a known platform. `visionOS` is the correct name.

PR #1196, opened 2023-12-14
Stringify `FormState` values in `<LiveForm>` elements

carson-katri:
This improves on the changes made in #1180:

1. `<LiveForm>` values will now be stringified
2. Existing strings will not be wrapped in a nested string
3. `FragmentEncoder` was removed as it is no longer necessary


simlay (approved):
👍

PR #1198, opened 2023-12-14
Fix cached page rendering when nested in `<NavigationStack>`

carson-katri:
This improves the cached page rendering when using `<NavigationStack>` in a layout.

Previously, the page would go blank during transitions. With these changes, the previous page will display with the `disabled` modifier applied until is focused. Here's a demo with animations slowed:

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/782db1d3-6db4-424a-9ebc-be2c1e489f84


simlay (approved):
I've regularly wondered when people use the slow rendering feature of the iOS simulator.

Issue #1199, opened 2023-12-16
How to use `NavigationLink` with `v0.2.0-beta.1`


Fudoshiki:
live_view_native: v0.2.0-beta.1
liveview_client_swiftui: v0.2.0-beta.1

**ContentView.swift**
```swift
import SwiftUI
import LiveViewNative

struct MyRegistry: RootRegistry {
    static func loadingView(for url: URL, state: LiveSessionState) -> some View {
        ProgressView()
    }
}

struct ContentView: View {
    @StateObject private var coordinator = LiveSessionCoordinator<MyRegistry>(
        .automatic(URL(string: "https://sharevari.com/")!),
        config: LiveSessionConfiguration()
    )

    var body: some View {
        LiveView(session: coordinator)
            .onReceive(coordinator.$state) { state in
                if case .connectionFailed = state {
                    Task {
                        // Delay of 5 seconds (1 second = 1_000_000_000 nanoseconds)
                        try await Task.sleep(nanoseconds: 5_000_000_000)
                        await coordinator.reconnect()
                    }
                }
            }
    }
}
```

**hello_live.swiftui.heex**
```elixir
<VStack id="hello-native">
  <Text>
    Hello native!
  </Text>
  <NavigationLink destination="/test">
    <HStack>
      <Text>
        Test
      </Text>
    </HStack>
  </NavigationLink>
</VStack>
```

**text_live.swiftui.heex**
```elixir
<VStack id="test-native">
  <Text>
    Hello native test!
  </Text>
</VStack>
```

**router.ex**
```elixir
pipeline :browser do
  plug LiveViewNative.SessionPlug
end

live "/test", TestLive
...
```

## Result

Blank page without request to `/test` in console

![simulator_screenshot_26DA155F-B467-4BD5-8954-388BC5304E00](https://github.com/liveview-native/liveview-client-swiftui/assets/4083992/8e32e4b2-6f25-4119-92f6-fd3a97cc98a7)

PR #1201, opened 2023-12-19
Connect `LiveViewCoordinator` from `NavigationLink`

carson-katri:
When `<NavigationLink>` was used, the coordinator was never able to connect.

This resolves the issue by handling coordinator connection on forward navigation from the `LiveSessionCoordinator`.


bcardarella:
I assume this closes #1199 ?

Sources/LiveViewNative/LiveView.swift
```diff
@@ -45,18 +45,15 @@ public macro LiveView<Host: LiveViewHost>(
...
     
     @Environment(\.scenePhase) private var scenePhase
     
     /// Creates a new LiveView attached to the given coordinator.
     ///
     /// - Note: Changing coordinators after the `LiveView` is setup and connected is forbidden.
-    public init(session: LiveSessionCoordinator<R>) {
-        self._session = .init(wrappedValue: session)
-        self.rootCoordinator = session.navigationPath.first!.coordinator
+    public init(session: @autoclosure @escaping () -> LiveSessionCoordinator<R>) {
```
carson-katri, line 55: I've made this take a closure instead, which should help with embedding a `LiveView` in an app with state that changes outside.

It's marked `@autoclosure` so the API looks the same as before.

Sources/LiveViewNative/LiveView.swift
```diff
@@ -66,6 +63,10 @@ public struct LiveView<R: RootRegistry>: View {
     public init(url: URL, configuration: LiveSessionConfiguration = .init()) {
         self.init(session: .init(url, config: configuration))
     }
+    
+    private var rootCoordinator: LiveViewCoordinator<R> {
+        session.navigationPath.first!.coordinator
+    }
```
carson-katri, line 69: The `rootCoordinator` is now a computed property. This allows `redirect` navigation to be performed from the root of the app.

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -364,19 +364,21 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
...
                 if !navigationPath.isEmpty {
-                    navigationPath[navigationPath.count - 1] = entry
+                    await navigationPath.last?.coordinator.disconnect()
+                    let coordinator = LiveViewCoordinator(session: self, url: redirect.to)
+                    coordinator.url = redirect.to
+                    navigationPath[navigationPath.count - 1] = LiveNavigationEntry(url: redirect.to, coordinator: coordinator)
+                    try await coordinator.connect(domValues: self.domValues, redirect: true)
                 }
             }
-            try await coordinator.connect(domValues: self.domValues, redirect: true)
```
carson-katri, line 374: Connection is now handled from `scan` in `LiveSessionCoordinator.init`.

Only `replace` navigation needs to manage its own reconnection.

carson-katri:
Here's an example of the navigation using a combination of `<NavigationLink>`, `push_navigate` and `push_navigate(replace: true)`:

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/ed492c22-7551-4279-b22a-b1473d74f83d

supernintendo (approved):

PR #1202, opened 2023-12-19
Use `supportedDestinations` to create a multiplatform target

carson-katri:
This makes a few tweaks to the generated Xcode project:

1. The route is passed to the `development` configuration as well as `production`:

```swift
//
//  ContentView.swift
//  XcodegenTestProject
//

import SwiftUI
import LiveViewNative

struct ContentView: View {
    var body: some View {
        LiveView(.automatic(
            development: .localhost(path: "/my-native-route"),
            production: .custom(URL(string: "https://myapp.com/my-native-route")!)
        ))
    }
}
```

2. A `tvOS (Experimental)` option is added to the list of possible targets. It is marked `Experimental` due to missing simulator support.

3. A multiplatform target is created for iOS, macOS, and tvOS platforms. This replaces the individual platform targets that were created before. watchOS remains a separate target.

Screenshot 2023-12-19 at 12 24 55 PM

Screenshot 2023-12-19 at 12 25 03 PM

4. Added a `.gitignore` to the generated files.


lib/mix/tasks/install.ex
```diff
@@ -136,14 +136,16 @@ defmodule Mix.Tasks.Lvn.SwiftUI.Install do
...
+    lvn_ios = if "iOS" in targets, do: "iOS", else: ""
+    lvn_macos = if "macOS" in targets, do: "macOS", else: ""
+    lvn_tvos = if "tvOS (Experimental)" in targets, do: "tvOS", else: ""
     lvn_watchos_include_path = if "watchOS" in targets, do: "project_watchos.yml", else: "skip_spec.yml"
     xcodegen_env = [
       {"LVN_APP_NAME", app_namespace},
-      {"LVN_BUNDLE_IDENTIFIER", "com.myorg.#{app_namespace}"},
-      {"LVN_IOS_INCLUDE_PATH", lvn_ios_include_path},
-      {"LVN_MACOS_INCLUDE_PATH", lvn_macos_include_path},
+      {"LVN_BUNDLE_IDENTIFIER", "com.example.#{app_namespace}"},
```
carson-katri, line 145: I've seen Apple recommend the `com.example.*` naming, so I changed this. It might be better to make this configurable during setup though, or infer it from the base URL.

supernintendo (approved):

PR #1203, opened 2023-12-20
Support custom modifiers from registries and in `ContentBuilder`

carson-katri:
This PR deals with modifiers in multiple contexts which I'll refer to as:
* "root context" - the main app, uses SwiftUI Views
* "content builder context" - custom element types in a specific View. For example, `MapContent` or `ChartContent` instead of View types.

Changes:

* `AggregateRegistry` now properly collects modifiers from addons. This also fixes the `#LiveView(_:addons:)` macro, which used `AggregateRegistry` internally.
* Custom modifiers from the registry will now be parsed after built-in modifiers (were previously ignored).
* `ContentBuilder` can declare custom modifiers and use classes on its elements. These should also be included in the `CustomRegistry` for the add-on to avoid parsing errors.
* `LiveViewNativeStylesheet` was made available to external packages for use in add-ons.
* `@ContentBuilderContext` now accepts the `ContentBuilder` type as a second generic argument. This is necessary for efficient stylesheet parsing.

> [!NOTE]
> Content builders parse stylesheets separate from the rest of the app. This is done on-demand, and does not impact screens without them.

Here's an example of using custom modifiers both in the root context, and in a content builder context:

```ex
"marker-blue" do
  tint(.blue) # uses the `MapContentBuilder` tint modifier when in that context.
end

"look-around" do
  # custom modifier that applies to Views
  lookAroundViewer(isPresented: attr("is-on"), initialScene: CLLocationCoordinate2D(latitude: 0, longitude: 0))
end
```

```html
<Map>
  <!-- children of `Map` use the `MapContentBuilder` -->
  <Marker class="marker-blue" latitude={38.8951} longitude={-77.0364}>
    Washington, D.C.
  </Marker>
</Map>
<Toggle phx-change="toggle-changed" is-on={@toggle} class="look-around">
  Show Look Around Viewer
</Toggle>
```

Here's what a modifier in a `ContentBuilder` looks like:

### Declare the modifier type
```swift
import SwiftUI
import MapKit
import LiveViewNative
import LiveViewNativeStylesheet

@ParseableExpression
struct TintModifier: ContentModifier {
    typealias Builder = MapContentBuilder
    
    static let name = "tint"
    
    let color: Color
    
    init(_ color: Color) {
        self.color = color
    }
    
    func apply<R: RootRegistry>(
        to content: Builder.Content,
        on element: ElementNode,
        in context: Builder.Context<R>
    ) -> Builder.Content {
        content.tint(color)
    }
}
```

> [!NOTE]
> Name collisions with built-in modifiers are ok in a `ContentBuilder`. In the root context, the built-in modifiers are always given priority. In a content builder's context, any built-in modifiers will be ignored.

### Use an enum in the `ContentBuilder` to parse all supported modifiers
```swift
enum MapContentBuilder: ContentBuilder {
    enum ModifierType: ContentModifier {
        typealias Builder = MapContentBuilder
        
        case tint(TintModifier)
        // ...
        
        static func parser(in context: ParseableModifierContext) -> some Parser<Substring.UTF8View, Self> {
            OneOf {
                TintModifier.parser(in: context).map(Self.tint)
                // ...
            }
        }
        
        func apply<R>(
            to content: Builder.Content,
            on element: ElementNode,
            in context: Builder.Context<R>
        ) -> Builder.Content where R : RootRegistry {
            switch self {
            case .tint(let modifier):
                return modifier.apply(to: content, on: element, in: context)
                // ...
            }
        }
    }

    // ...
}
```

### Add to the `CustomRegistry`
It is recommended to declare content builder modifiers in the `CustomRegistry` as well. This will avoid parsing errors, and can catch errors the content builder might miss.
```swift
import LiveViewNative
import LiveViewNativeStylesheet
import SwiftUI

public enum MapKitRegistry<Root: RootRegistry>: CustomRegistry {
    // ...

    public struct CustomModifier: ViewModifier, ParseableModifierValue {
        enum Storage {
            // ...
            case noop
        }
        
        let storage: Storage
        
        public static func parser(in context: ParseableModifierContext) -> some Parser<Substring.UTF8View, Self> {
            OneOf {
                // ...
                // This line will catch any ContentBuilder modifiers, and map them to a `noop` type.
                MapContentBuilder.ModifierType.parser(in: context).map({ _ in Self(storage: .noop) })
            }
        }
        
        public func body(content: Content) -> some View {
            switch storage {
            // ...
            case .noop:
                content
            }
        }
    }
}
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,6 +17,9 @@ let package = Package(
         .library(
             name: "LiveViewNative",
             targets: ["LiveViewNative"]),
+        .library(
+            name: "LiveViewNativeStylesheet",
+            targets: ["LiveViewNativeStylesheet"]),
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
--- a/Sources/LiveViewNative/ContentBuilder.swift
+++ b/Sources/LiveViewNative/ContentBuilder.swift
@@ -5,8 +5,9 @@
 //  Created by Carson Katri on 6/15/23.
 //
 
-import LiveViewNativeCore
 import SwiftUI
+import LiveViewNativeCore
+import LiveViewNativeStylesheet
 
 /// A protocol add-on libraries conform to support custom DSLs.
 ///
@@ -187,11 +188,6 @@ public protocol ContentBuilder {
     /// An enum with the supported element names.
     associatedtype TagName: RawRepresentable<String>
     
-    /// A `Decodable` enum with the supported modifier names.
-    ///
-    /// The raw values for a modifier should be in `snake_case`.
-    associatedtype ModifierType: RawRepresentable<String>, Decodable
-    
     /// The type of content this builder produces.
     ///
     /// A protocol cannot be used as the content type.
@@ -233,6 +229,8 @@ public protocol ContentBuilder {
     /// ```
     associatedtype Content
     
+    associatedtype ModifierType: ContentModifier = EmptyContentModifier<Self> where ModifierType.Builder == Self
+    
     /// Switch over the ``tag`` and provide the appropriate ``Content``.
     ///
     /// The ``ObservedElement`` and ``LiveContext`` property wrappers cannot be used outside of `View`.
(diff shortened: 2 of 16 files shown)


carson-katri:
You can see the diff for MapKit to get a general overview of how the API changes affect addons: https://github.com/liveview-native/liveview-native-swiftui-mapkit/pull/206/files

bcardarella (approved):

PR #1214, opened 2024-01-03
Support multi-item selection

### LINKED ISSUE #1213: Support multiple List row selections
Currently, you can only select one item at a time with a list. This should be changed to allow multiple elements as well. One way to do this would be allowing a JSON encoded array of string IDs to be passed to `selection`.

carson-katri:
Closes #1213 

Pass a JSON encoded array to allow for multi-select. This also works with the `EditButton` element.

```ex
def mount(_params, _session, socket) do
  {:ok, assign(socket, selection: [])}
end

def render(%{format: :swiftui} = assigns) do
  ~SWIFTUI"""
  <EditButton />
  <List selection={Jason.encode!(@selection)} phx-change="selection-changed">
    <Text id={"#{i}"} :for={i <- 1..10}><%= i %></Text>
  </List>
  """
end

def handle_event("selection-changed", %{ "selection" => selection }, socket) do
  {:noreply, assign(socket, selection: selection)}
end
```

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/c9ac86fc-a8d9-43c3-bb2f-af951620482e

### DIFF
--- a/Sources/LiveViewNative/Utils/Selection.swift
+++ b/Sources/LiveViewNative/Utils/Selection.swift
@@ -5,6 +5,7 @@
 //  Created by Carson Katri on 2/21/23.
 //
 
+import Foundation
 import LiveViewNativeCore
 
 /// Selection of either a single value or set of values.
@@ -59,10 +60,14 @@ enum Selection: Codable, AttributeDecodable, Equatable {
     init(from attribute: LiveViewNativeCore.Attribute?) throws {
         guard let value = attribute?.value
         else { throw AttributeDecodingError.missingAttribute(Self.self) }
-        if value.isEmpty {
-            self = .single(nil)
-        } else {
-            self = .single(value)
+        do {
+            self = try JSONDecoder().decode(Self.self, from: Data(value.utf8))
+        } catch {
+            if value.isEmpty {
+                self = .none
+            } else {
+                self = .single(value)
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


carson-katri:
@bcardarella Is it possible to have the LVN engine automatically JSON encode list attributes?

Do you have an example of the Phoenix select options I could look at?

bcardarella:
We could but let me first find out if there are any traps with doing this

BrooklinJazz:
> @bcardarella Is it possible to have the LVN engine automatically JSON encode list attributes?
> 
> Do you have an example of the Phoenix select options I could look at?

@carson-katri in the `.input` core component, this function converts a keyword list and a select value into the options for a select element.
https://hexdocs.pm/phoenix_html/3.3.3/Phoenix.HTML.Form.html#options_for_select/2

This is how it's defined in core components:

```elixir
  def input(%{type: "select"} = assigns) do
    ~H"""
    <div phx-feedback-for={@name}>
      <.label for={@id}><%= @label %></.label>
      <select
        id={@id}
        name={@name}
        class="mt-2 block w-full rounded-md border border-gray-300 bg-white shadow-sm focus:border-zinc-400 focus:ring-0 sm:text-sm"
        multiple={@multiple}
        {@rest}
      >
        <option :if={@prompt} value=""><%= @prompt %></option>
        <%= Phoenix.HTML.Form.options_for_select(@options, @value) %>
      </select>
      <.error :for={msg <- @errors}><%= msg %></.error>
    </div>
    """
  end
```

Meaning options can be defined like so:

```elixir
<.input type="select" options={["label": "value", "label2": "value2"]}/>
```

Hope that gives some inspiration!

AZholtkevych:
@carson-katri could you please merge it if it is finished? Thank you so much

PR #1233, opened 2024-01-18
Support non-key-value lists

### LINKED ISSUE #1231: Rules parser does not support un-labelled lists
Passing a list of static members does not compile:

```swift
Gradient(colors: [.red, .blue])
```

fails with:

```
== Compilation error in file lib/lvn_demo3_web/styles/home_styles.ex ==
** (SyntaxError) :1: Unsupported input:
  |
1 | Gradient(colors: [.red, .blue])
  |                  ^^^^^^
  |

Expected one of the following:
 - a keyword list eg ‘[style: :dashed]’, ‘[size: 12]’ or ‘[lineWidth: lineWidth]’
 - a Swift range eg ‘1..<10’ or ‘foo(Foo.bar...Baz.qux)’
 - a number, string, nil, boolean or :atom
 - an event eg ‘event("search-event", throttle: 10_000)’
 - an attribute eg ‘attr("placeholder")’
 - an IME eg ‘Color.red’ or ‘.largeTitle’ or ‘Color.to_ime(variable)’
 - a helper function eg ‘to_float(variable)’
 - a modifier eg ‘bold()’
 - a variable defined in the class header eg ‘color_name’
```

The same error occurs with the following inputs:
```swift
Gradient(colors: [Color.red, Color.blue])
Gradient(colors: ["red", "blue"])
Gradient(colors: [0, 1, 2])
```

Keyword lists with index keys can be used in the meantime:
```swift
Gradient(colors: [0: .red, 1: .blue])
```

NduatiK:
Resolves #1231 

@carson-katri, please confirm that the ASTs are what you'd expect.

### DIFF
--- a/lib/live_view_native/swiftui/rules_parser/expressions.ex
+++ b/lib/live_view_native/swiftui/rules_parser/expressions.ex
@@ -6,11 +6,13 @@ defmodule LiveViewNative.SwiftUI.RulesParser.Expressions do
 
   def enclosed(start \\ empty(), open, combinator, close, opts) do
     allow_empty? = Keyword.get(opts, :allow_empty?, true)
+    generate_error? = Keyword.get(opts, :generate_error?, true)
 
     close =
       expect(
         ignore(string(close)),
         error_message: "expected ‘#{close}’",
+        generate_error?: generate_error?,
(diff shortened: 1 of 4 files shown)


carson-katri (approved):
LGTM 👍

PR #1234, opened 2024-01-18
Customizable `LiveView` state views

### LINKED ISSUE #1232: Customizable `LiveView` state indicator Views
Currently, customizing the loading, error, etc. states of a `LiveView` requires a `RootRegistry`. This API could be simplified by providing extra `ViewBuilder` closures on the `LiveView` initializer:

```swift
LiveView(.localhost) {
  ProgressView("Loading")
} error: { error in
  #if DEBUG
  Text(error.localizedDescription)
  #else
  Text("An error occurred")
  #endif
} reconnecting: { content in
  content
    .overlay {
      Text("Reconnecting")
    }
} disconnected: {
  Text("Failed to connect")
}
```

These arguments would be given priority over the `RootRegistry`. If they are not provided, the Views in the `RootRegistry` will be used.

carson-katri:
Closes #1232 

State views can be provided as arguments to the `LiveView`. The `loadingView` method was from the `CustomRegistry` protocol, as it makes more sense to provide these options on the `LiveView` itself. Here's an example of using these arguments:

```swift
LiveView(.localhost) {
    ProgressView("Loading")
} disconnected: {
    Text("Not connected to the server")
} reconnecting: { content in
    content
        .safeAreaInset(edge: .bottom) {
            HStack {
                ProgressView()
                Text("Reconnecting")
            }
            .padding()
            .frame(maxWidth: .infinity, alignment: .leading)
            .background(Color.orange)
        }
} error: { error in
    Text(error.localizedDescription)
}
```

Note that `reconnecting` provides the `content` of the `LiveView`, allowing you to create an overlay instead of fully replacing the screen when the connection temporarily drops.

All of these arguments are optional. If any is omitted, a default View will be used. They can also be used with the `#LiveView` macro when passing `addons`.

A new `registry` argument was also added. Previously, registries were passed as generic arguments:
```swift
LiveView<MyRegistryType>(.localhost)
```
With the addition of these state views, there are more generic arguments that must be provided.
```swift
LiveView<MyRegistryType, _, _, _, _>(.localhost)
```
Instead of passing a generic argument, it is now recommended to pass the `registry` argument:
```swift
LiveView(registry: MyRegistryType.self, .localhost)
```
However, it is generally easier to use the `#LiveView` macro to pass custom registries, so this argument is almost never required unless some further customization is required:
```swift
#LiveView(.localhost, addons: [MapKitRegistry<_>.self, ChartsRegistry<_>.self])
```

### DIFF
--- a/Sources/LiveViewNative/NavStackEntryView.swift
+++ b/Sources/LiveViewNative/NavStackEntryView.swift
@@ -43,23 +43,19 @@ struct NavStackEntryView<R: RootRegistry>: View {
                         .id(ObjectIdentifier(document))
                 } else {
                     SwiftUI.Group {
-                        if R.LoadingView.self == Never.self {
-                            switch coordinator.state {
-                            case .connected:
-                                fatalError()
-                            case .notConnected:
-                                SwiftUI.Text("Not Connected")
-                            case .connecting:
-                                SwiftUI.ProgressView("Connecting")
-                            case .connectionFailed(let error):
-                                SwiftUI.VStack {
-                                    SwiftUI.Text("Connection Failed")
-                                        .font(.subheadline)
-                                    SwiftUI.Text(error.localizedDescription)
-                                }
+                        switch coordinator.state {
+                        case .connected:
+                            fatalError()
+                        case .notConnected:
+                            SwiftUI.Text("Not Connected")
+                        case .connecting:
+                            SwiftUI.ProgressView("Connecting")
+                        case .connectionFailed(let error):
+                            SwiftUI.VStack {
+                                SwiftUI.Text("Connection Failed")
+                                    .font(.subheadline)
+                                SwiftUI.Text(error.localizedDescription)
                             }
-                        } else {
-                            R.loadingView(for: coordinator.url, state: coordinator.state)
                         }
                     }
                     .transition(coordinator.session.configuration.transition ?? .identity)
--- a/Sources/LiveViewNative/NavStackEntryView.swift
+++ b/Sources/LiveViewNative/NavStackEntryView.swift
@@ -12,6 +12,7 @@ struct NavStackEntryView<R: RootRegistry>: View {
     private let entry: LiveNavigationEntry<R>
     @ObservedObject private var coordinator: LiveViewCoordinator<R>
     @StateObject private var liveViewModel = LiveViewModel()
+    @Environment(\.liveViewStateViews) private var liveViewStateViews
     
     init(_ entry: LiveNavigationEntry<R>) {
         self.entry = entry
@@ -47,15 +48,11 @@ struct NavStackEntryView<R: RootRegistry>: View {
                         case .connected, .reconnecting:
                             fatalError()
                         case .notConnected:
-                            SwiftUI.Text("Not Connected")
+                            liveViewStateViews.disconnectedView()
                         case .connecting:
-                            SwiftUI.ProgressView("Connecting")
+                            liveViewStateViews.connectingView()
                         case .connectionFailed(let error):
-                            SwiftUI.VStack {
-                                SwiftUI.Text("Connection Failed")
-                                    .font(.subheadline)
-                                SwiftUI.Text(error.localizedDescription)
-                            }
+                            liveViewStateViews.errorView(error)
                         }
                     }
                     .transition(coordinator.session.configuration.transition ?? .identity)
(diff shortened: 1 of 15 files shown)


bcardarella:
I'm going to go through the js code and just verify this represents all of the states we could go into

carson-katri:
`reconnect` states can now be rendered:

```swift
LiveView(.localhost) {
    ProgressView("Loading")
} reconnecting: { content in
    content
        .safeAreaInset(edge: .bottom) {
            HStack {
                ProgressView()
                Text("Reconnecting")
            }
            .padding()
            .frame(maxWidth: .infinity, alignment: .leading)
            .background(Color.orange)
        }
}
```

In this video, I killed the server while the app was running, then started it again a few seconds later:

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/f9f4fca8-8656-4760-8844-e4048aaa6fa5

carson-katri:
A default `reconnecting` is provided:

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/8fcb9fa1-e699-4565-845d-758977faf67a


https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/484a1d65-31d4-49fd-a3c1-2a7323137510

bcardarella:
This is excellent!

bcardarella (approved):

Sources/LiveViewNative/NavStackEntryView.swift
```diff
@@ -43,23 +43,19 @@ struct NavStackEntryView<R: RootRegistry>: View {
...
+                        case .connected, .reconnecting:
+                            fatalError()
+                        case .notConnected:
+                            SwiftUI.Text("Not Connected")
+                        case .connecting:
+                            SwiftUI.ProgressView("Connecting")
+                        case .connectionFailed(let error):
+                            SwiftUI.VStack {
+                                SwiftUI.Text("Connection Failed")
+                                    .font(.subheadline)
```
bcardarella, line 56: I think we should remove this from being default, and that all default states should be `nil`. Instead we should move these into the template project generator so people can have direct access to modify.

bcardarella (changes requested):

Sources/LiveViewNative/NavStackEntryView.swift
carson-katri: So should the state view arguments be required? Or do states like `connecting`, `disconnected`, etc. just render a blank screen if no argument is passed?

bcardarella: No, not required. If they aren't provided it's just a `no-op`

carson-katri: How about the LiveView error webpage View. Should we display that by default?

bcardarella: oh that's a good point, how about the view is shipped with the client but we still have the code to render that view in the error closure generated into the template

carson-katri:
@bcardarella Here are the final template states:

<table>
<tr><th>Connecting</th><th>Reconnecting</th><th>Error (Release Mode)</th><th>Server Error (Debug Mode)</th><th>Connection Error (Debug Mode)</th></tr>
<tr>
<td>



</td>
<td>



</td>
<td>

Simulator Screenshot - iPhone 15 Pro - 2024-01-31 at 13 57 06

</td>
<td>

Simulator Screenshot - iPhone 15 Pro - 2024-02-01 at 14 42 26

</td>
<td>

Simulator Screenshot - iPhone 15 Pro - 2024-02-01 at 14 40 29

</td>
</tr>
</table>

priv/xcodegen/Sources/TemplateApp/ContentView.swift
```diff
@@ -8,9 +8,85 @@ import LiveViewNative
...
-            development: .localhost(path: "%LVN_PREFERRED_ROUTE%"),
-            production: .custom(URL(string: "https://%LVN_PREFERRED_PROD_URL%%LVN_PREFERRED_ROUTE%")!)
-        ))
+        #LiveView(
+            .automatic(
+                development: .localhost(path: "%LVN_PREFERRED_ROUTE%"),
+                production: URL(string: "https://%LVN_PREFERRED_PROD_URL%%LVN_PREFERRED_ROUTE%")!
+            ),
+            addons: []
+        ) {
```
carson-katri, line 17: The template uses the `#LiveView` macro with an empty `addons` parameter by default now.

bcardarella (approved):

PR #1246, opened 2024-01-25
Refactor `@ParseableExpression` macro for easier debugging

carson-katri:
The `ParseableExpressionMacro` was using swift-parsing's declarative API, which made debugging the output more difficult. This PR refactors it to use an imperative style, which is easier to work with in a macro.


bcardarella:
I just tried the branch. I'm getting this error with the modifier parser now:

```
Stylesheet parsing failed for modifier `clipShape` in class `clipShape:circle`:

  |
5 | clipShape(.circle)
  | ^ Unknown modifier `clipShape`

in BostonElixirWeb.Styles.SwiftUI (swfitui.ex:5)
```

carson-katri:
The bug found with `clipShape` should be resolved.

bcardarella:
confirmed this is now fixed

bcardarella (approved):

PR #1248, opened 2024-01-31
Display dead render until connected

### LINKED ISSUE #1229: Display the dead render while the socket connects
Currently we display a loading indicator while the socket connects. According to @carson-katri Now that we have platform-specific dead renders, we could now display the actual page content instead (likely with the disabled modifier applied to the views until the socket connects.

This would improve our initial page load time.

carson-katri:
Closes #1229 

The dead render is displayed with the `disabled` modifier, so no interaction is permitted until the channel connection is established.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -154,6 +154,8 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
             
             self.domValues = domValues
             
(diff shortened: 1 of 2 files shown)


bcardarella:
I'm actually :-1: on this feature.

* I'd like to determine first how much additional time establishing the LV connection has, on the web it is very fast to hydrate. * the dead render UI would be read-only and it could lead to a confusing user experience if users are expecting to click on UI elements. 
* there's a larger issue of state management that we've yet to tackle. For example, if you are logged in but now a dead render could be rendering a non-auth state then LV connects and it's an auth state

bcardarella:
Closing as #1229 is invalid

PR #1250, opened 2024-02-01
Lookup manually-implemented types during schema generation

carson-katri:
Closes https://github.com/liveview-native/liveview-native-vscode/issues/17


bcardarella:
@carson-katri will this need to be done on each new version of swift? And I assume this won't work for modifiers/views that are part of other libs?

carson-katri:
I marked this as draft because I'm refactoring the schema generator to:
1. Generate the list automatically from the project source
2. Move the schema from the VS Code extension into this library
3. Allowing it to generate for arbitrary directories (to work with other libraries)

carson-katri:
The generator now produces a JSON file based on a `interface`, `modifier-search-path` and `enum-search-path`. The schema file is stored at `stylesheet-language.json`. `liveview-native-vscode` can look for this file in all of the open workspace's dependencies to create a full list of completions.

carson-katri:
1. As it is implemented now, they will need to run the generator with `swift run ModifierGenerator --modifier-search-path --enum-search-path` and save the output to a file called `stylesheet-language.json` at the root of the addon repo.
2. Yes, other platforms could use `stylesheet-language.json`. I gave it a `format` key so it could in theory switch between them. But the VS Code extension does not currently handle other types of stylesheet.

bcardarella:
@carson-katri and I spoke today, we'll need to think about where the `stylesheet-language.json` file will live. The currently implementation assumed `_deps/**/*/stylesheet-language.json` as the glob to find (or a pattern similar) but there are LVN addons that don't aren't Elixir deps. Now we are introducing platform specific searching into the vscode plugin. Instead there should probably be a host that allows the storing and retrieval of all the language files. But, do we have an all or nothing or do we want to provide a way to just install what you want or what you need?

carson-katri:
I've added a `list` subcommand, which will output a newline-separated list of all available modifiers. This can be used with #1294 

```
OVERVIEW: Output a list of the names of all available modifiers.

USAGE: modifier-generator list [--interface <interface>] [--modifier-search-path <modifier-search-path>]

OPTIONS:
  --interface <interface> The `.swiftinterface` file from
                          `/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/SwiftUI.framework/Modules/SwiftUI.swiftmodule/arm64-apple-ios.swiftinterface`
  --modifier-search-path <modifier-search-path>
                          The path to search for additional modifier types
  -h, --help              Show help information.
```

PR #1252, opened 2024-02-02
Extract state views into their own views and address warnings

bcardarella:
The state views are extracted into their own views to make the ContentView less cluttered and to provide a clean location to make changes. This also allows Previews to be used.

Also suppress the recommended changes warnings when generating a new project

### DIFF
--- a/priv/xcodegen/Sources/TemplateApp/ContentView.swift
+++ b/priv/xcodegen/Sources/TemplateApp/ContentView.swift
@@ -15,78 +15,14 @@ struct ContentView: View {
             ),
             addons: []
         ) {
+          // connecting
             ProgressView()
         } disconnected: {
-            if #available(iOS 17, macOS 14, tvOS 17, watchOS 10, *) {
-                ContentUnavailableView {
-                    Label("No Connection", systemImage: "network.slash")
-                } description: {
-                    Text("The app will reconnect when network connection is regained.")
-                }
-            } else {
-                VStack {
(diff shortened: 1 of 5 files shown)


bcardarella:
@carson-katri I'd like to get this in for alpha.4 release today. I know you're PTO so I'll merge as this is working and if there are changes you'd like to see we can discuss those next week when you're available

priv/xcodegen/Sources/TemplateApp/ReconnectingView.swift
```diff
@@ -0,0 +1,42 @@
...
+                        .transition(.move(edge: .top).combined(with: .opacity))
+                }
+            }
+            .animation(.default, value: isReconnecting)
+    }
+}
+
+//#Preview {
+//    ReconnectingView(
+//        content: Text("Test"),
```
bcardarella, line 39: @carson-katri for this I was getting an error about how `Text` cannot be converted to `AnyView`. I thought that `AnyView` would match on any view struct?

Screenshot 2024-02-02 at 9 09 15 AM

priv/xcodegen/Sources/TemplateApp/ContentView.swift
```diff
@@ -15,78 +15,14 @@ struct ContentView: View {
             ),
             addons: []
         ) {
+          // connecting
```
bcardarella, line 18: because Swift doesn't support labels for the default closure I think this comment at least implies its purpose for those that are not familiar with Swift

PR #1258, opened 2024-02-06
Safely unwrap `NavigationLink` and `Link` URLs

### LINKED ISSUE #1256: NavigationLink view fatal error when no url provided
```heex
<NavigationLink>
  ...
</NavigationLink>
```

results in this:

Screenshot 2024-02-02 at 10 52 45 AM

In the event of a missing `url` attribute it should just render and not error like this.

carson-katri:
Closes #1256 

If the URL is invalid, the content of the View will be rendered instead.

### DIFF
--- a/Sources/LiveViewNative/Views/Controls and Indicators/Links/Link.swift
+++ b/Sources/LiveViewNative/Views/Controls and Indicators/Links/Link.swift
@@ -6,6 +6,9 @@
 //
 
 import SwiftUI
+import OSLog
+
+private let logger = Logger(subsystem: "LiveViewNative", category: "Link")
 
 /// Opens a URL when tapped.
 ///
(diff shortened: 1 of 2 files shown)


bcardarella:
I agree with the unwrapping in the event of missing arguments. I'm wondering though if we should also log a warning?

carson-katri:
This will now log when missing the `destination` attribute:
```
Missing or invalid `destination` on `<Link>Test</Link>`
Missing or invalid `destination` on `<NavigationLink>Test</NavigationLink>`
```

bcardarella:
@carson-katri if I'm understanding correctly the unwrap will just render the content and hoist? Does that mean if any modifiers are applied to `Link` and `NavigationLink` that are meant to apply to child views will be lost? If that's the case can we just replace with `Group` as the container, copy the modifiers over, and not hoist?

carson-katri:
Modifiers are still applied.

bcardarella (approved):

PR #1260, opened 2024-02-06
Update template project Views

### LINKED ISSUE #1253: Reconnecting state view is rendering in the middle
Screenshot 2024-02-02 at 9 59 44 AM

LVN Markup:

```heex
<VStack class="padding" alignment="leading">
  <Text class="font:title">Turtle Rock</Text>
  <HStack class="font:subheadline">
    <Text>Joshua Tree National Park</Text>
    <Spacer />
    <Text>California</Text>
  </HStack>
</VStack>
```

carson-katri:
Closes #1253 

This makes some refinements to the template Views. This includes adding a flexible `frame` modifier to the content of a `ReconnectingView` to ensure the banner appears at the top of the screen.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveConnectionError.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveConnectionError.swift
@@ -53,7 +53,7 @@ public enum LiveConnectionError: Error, LocalizedError {
             case .document:
                 return "document"
             case .csrfToken:
-                return #"<meta name="csrf-token">"#
+                return #"<.csrf_token />"#
             case .phxMain:
                 return #"<div data-phx-main="...">"#
             }
--- a/priv/xcodegen/Sources/TemplateApp/ContentView.swift
+++ b/priv/xcodegen/Sources/TemplateApp/ContentView.swift
@@ -15,8 +15,7 @@ struct ContentView: View {
             ),
             addons: []
         ) {
-            // connecting
-            ProgressView()
+            ConnectingView()
         } disconnected: {
             DisconnectedView()
         } reconnecting: { content, isReconnecting in
--- a/priv/xcodegen/Sources/TemplateApp/ContentView.swift
+++ b/priv/xcodegen/Sources/TemplateApp/ContentView.swift
@@ -15,12 +15,14 @@ struct ContentView: View {
             ),
             addons: []
         ) {
-          // connecting
+            // connecting
             ProgressView()
         } disconnected: {
             DisconnectedView()
         } reconnecting: { content, isReconnecting in
-            ReconnectingView(content: content, isReconnecting: isReconnecting)
+            ReconnectingView(isReconnecting: isReconnecting) {
+                content
+            }
         } error: { error in
             ErrorView(error: error)
         }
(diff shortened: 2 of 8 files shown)


Sources/LiveViewNative/Coordinators/LiveConnectionError.swift
```diff
@@ -53,7 +53,7 @@ public enum LiveConnectionError: Error, LocalizedError {
             case .document:
                 return "document"
             case .csrfToken:
-                return #"<meta name="csrf-token">"#
+                return #"<.csrf_token />"#
```
bcardarella, line 56: what is this doing?

priv/xcodegen/Sources/TemplateApp/ContentView.swift
```diff
@@ -15,14 +15,16 @@ struct ContentView: View {
...
             addons: []
         ) {
-          // connecting
+            // connecting
             ProgressView()
         } disconnected: {
             DisconnectedView()
         } reconnecting: { content, isReconnecting in
-            ReconnectingView(content: content, isReconnecting: isReconnecting)
+            ReconnectingView(isReconnecting: isReconnecting) {
```
bcardarella, line 23: now I'm curious why the original method didn't work

priv/xcodegen/Sources/TemplateApp/ContentView.swift
```diff
@@ -15,14 +15,16 @@ struct ContentView: View {
             ),
             addons: []
         ) {
-          // connecting
+            // connecting
```
bcardarella, line 18: I think we should add a `ConnectingView` that `ProgressView` will be put into.

Sources/LiveViewNative/Coordinators/LiveConnectionError.swift
carson-katri: This is used as part of the error message when the CSRF token is missing.

priv/xcodegen/Sources/TemplateApp/ContentView.swift
carson-katri: If you're referring to the banner not being at the top of the screen, that was fixed with this modifier: https://github.com/liveview-native/liveview-client-swiftui/pull/1260/files#diff-582b7403b42572a7b5f6ee58beec2c6d043fae407237f10da7d9b157d00cede3R14

bcardarella: I was curious why using the closure instead of passing the value to the named argument or is this just a style decision?

carson-katri: Its a style thing, but It also makes it easier to use in the preview because now it won't force you to erase the content to an `AnyView`.

carson-katri:
I've made some minor refactors to avoid type erasure in the `reconnecting` closure.

carson-katri:
Looking at this API again, I wonder if we should provide an option that passes an enum, similar to `AsyncImage`, that you can switch over with the state. This would provide more flexibility over animations/transitions. It would exist along with the multiple-closures API.

```swift
LiveView(.localhost) { phase in
  switch phase {
  case .connecting:
    ConnectingView()
  case let .connected(content):
    content
  case let .reconnecting(content):
    ReconnectingView {
      content
    }
  case let .failure(error):
    ErrorView(error: error)
  case .disconnected:
    DisconnectedView()
  }
}
```

bcardarella:
@carson-katri I defer to you on that, if you feel it is a beter API I'm :+1:

carson-katri:
I've added the `LiveViewPhase` closure API in 805883a
This also opens up the ability to apply modifiers to the rendered content of the `LiveView`. It should also make it easier to do more complex transitions between different phases.

bcardarella (approved):

PR #1261, opened 2024-02-06
ViewBuilder closure argument support

### LINKED ISSUE #1243: Proposal: Addressing closures
Example, in `AsyncImage` a content closure can be passed to inspect the phase state to render out if the images loads or there is some type of failure:

```swift
AsyncImage(url: URL(string: "https://example.com/icon.png")) { phase in
    if let image = phase.image {
        image // Displays the loaded image.
    } else if phase.error != nil {
        Color.red // Indicates an error.
    } else {
        Color.blue // Acts as a placeholder.
    }
}
```

we currently don't have a convention to manage this in LVN. Our own `AsyncImage` implementation hard codes these phases:

https://github.com/liveview-native/liveview-client-swiftui/blob/main/Sources/LiveViewNative/Views/Images/AsyncImage.swift#L49-L60

But we have a way to pass templates off to a closure. If `phase` is the conventional term used in these cases I'd like to propose the following:

```heex
<AsyncImage url={@url} >
  <Image class="foo-bar" phase="image" />
  <Text phase="error">Error!</Text>
</AsyncImage>
```

this convention should carry across all views

carson-katri:
Closes #1243

Current API:
```html
<AsyncImage url="...">
    <Group template="phase.empty">
        <ProgressView />
    </Group>
    <Group template="phase.success">
        <AsyncImage image class="resizable" /> <!-- reference to evaluated image element -->
    </Group>
    <Group template="phase.failure">
        <AsyncImage error /> <!-- reference to plain-text error description -->
    </Group>
</AsyncImage>
```

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
@@ -233,7 +233,6 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
         connectParams["_csrf_token"] = domValues.phxCSRFToken
         connectParams["_lvn"] = LiveSessionCoordinator<R>.platformParams
 
-        print("CONNECT LV:", self.url)
         let params: Payload = [
             "session": domValues.phxSession,
             "static": domValues.phxStatic,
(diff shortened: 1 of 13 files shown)


bcardarella:
@carson-katri can we go with my recommendation from #1243 with `template="phase[error]"` => `template="phase.error"` style?

PR #1262, opened 2024-02-06
Connect live reload socket when initial connection fails

### LINKED ISSUE #1255: Client doesn't attempt to reconnect if initial connection fails
Steps to reproduce:

1. Start client without a server running
2. Wait until the `Connection Failed` error renders
3. Start the server
4. Client doesn't attempt any reconnects

The expectation would be that the client is attempting reconnects indefinitely

carson-katri:
Closes #1255

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -233,6 +233,9 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
         } else {
             if let html = String(data: data, encoding: .utf8)
             {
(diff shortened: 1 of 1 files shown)


bcardarella:
@carson-katri I want to ensure we don't get a double connect here. If we have this error page load first and the WS connects, then the reload happens and we get a proper connection will the client establish a 2nd WS connection for live reload?

carson-katri:
No, there was already logic to disconnect if the socket is already setup: https://github.com/liveview-native/liveview-client-swiftui/blob/39c36d0b8e4e33723975a83a88e339ba891dd47f/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift#L348-L351

bcardarella (approved):

PR #1264, opened 2024-02-08
Restore navigation path after live reloading

### LINKED ISSUE #1263: LiveReload should reconnect on the current url
When navigating away from the base url in development if LiveReload is triggered the app reloads to the original base url. Instead it should reload on the current url navigated to

carson-katri:
Closes #1263 

This will still perform the reconnect on the original URL, but then perform a navigation back to the path when it lost connection.

I can change the implementation if it should instead do the dead render connection from the last URL in the path.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -183,8 +183,10 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
     ///
     /// This can be used to force the LiveView to reset, for example after an unrecoverable error occurs.
     public func reconnect() async {
+        let previousNavigationPath = self.navigationPath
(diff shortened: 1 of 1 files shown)


bcardarella:
@carson-katri I don't think this is the correct implementation. This will cause an unnecessary render of endpoint of the base url. What we want to do is instead use the query params as shown here: https://github.com/phoenixframework/phoenix_live_view/blob/main/assets/js/phoenix_live_view/view.js#L137-L147 so that there is no redirect that happens. It just does one live connection to the intended liveview.

carson-katri:
@bcardarella The LiveView will now reconnect from the topmost page in the navigation path. It also preserves the stale documents across reconnects, so navigation backward will be smoother after a reconnect.

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/03dcc54c-8f9b-47bf-a3c2-7a9d25f645bb

PR #1266, opened 2024-02-11
Remove guides

bcardarella:


carson-katri:
Is there a reason why the architecture documents were removed? I think they were valuable for contributors to understand how the client works.

bcardarella:
@carson-katri they're going to be re-written. The architecture has completely changed from v0.2.0. I started to re-write them but found they were requiring more time to re-write than write from scratch. However, if there are client/server architeecture guides that you feel are important we certainly should bring those back when I 'm re-writing the guides. They're in the git history so I didn't look at this as a "gone forever" thing

carson-katri:
Those documents described the architecture of the SwiftUI client, which is largely unchanged. For example, it described how modifier parsers are code-generated, the role of `LiveSessionCoordinator` and `LiveViewCoordinator`, the way View updates are handled, etc.

bcardarella:
Gotcha, maybe I included the wrong guides in the nuke then. I'll look to bring those back this week. Sorry about that and thanks for the catch!

bcardarella:
@AZholtkevych can you track this so I don't forget

bcardarella:
@carson-katri I'll cherry pick revert as the changes were caught up in a larger documentation change so I'd like to avoid just  a revert of this PR

PR #1270, opened 2024-02-14
Support `NavigationLink` for split view by creating separate socket connections

carson-katri:
Ideally, we would be able to re-use the socket. However, I was not able to find a way to initiate a separate LiveView from the client without the first LiveView being replaced/disconnected.

> [!NOTE]
> ~~If you put the `<NavigationSplitView>` in the root layout, it will be rendered in all columns. Either use a different layout for the secondary columns, or put the split view directly in the first column's template.~~
> [This has been changed](https://github.com/liveview-native/liveview-client-swiftui/pull/1270/commits/2c90ad48e50a6bc4b8ae243bcd900d2883f5b459) so if a `NavigationSplitView` is used within another `NavigationSplitView`, it will only render its sidebar content.

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/5ff103ad-b2ce-40d5-94e4-a0dbbf9e2404

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -48,7 +48,7 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
     // Socket connection
     var socket: Socket?
     
-    private var domValues: DOMValues!
+    private(set) var domValues: DOMValues!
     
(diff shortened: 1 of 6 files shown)


bcardarella:
I like this but let's see if we can grab some of Chris' time after I'm back from PTO to discuss the implementation here to ensure it aligns with web

bcardarella:
We're going to close this in favor of nested LiveViews which we will support in Core

PR #1285, opened 2024-02-29
Decode attributes with no values to empty strings

### LINKED ISSUE #1284: `String` attributes fail to decode when empty
When an empty `String` is used for an attribute, it fails decoding. An empty attribute should instead decode to an empty string.

This currently causes empty `ChangeTracked` string values to crash.

carson-katri:
Closes #1284


bcardarella:
@carson-katri how does this present? Because I think LiveView will take attributes like this:

```heex
<Text foobar={} />
```

and just drop the attribute

```heex
<Text />
```

I'll have to test it out but falsey values drop the attribute. Actually, I think I need to override this for @nelson-glauber as we want to preserver `false` for a value instead of dropping the value.

carson-katri:
When the attribute is fully dropped, it will still fail to decode (so you get proper "missing attribute" warnings). This is just for the cases where it would encode it like a boolean attribute `<Text value />` because the content was empty.

bcardarella:
gotcha ok

bcardarella (approved):

PR #1293, opened 2024-03-19
Fetch stylesheets from URL

### LINKED ISSUE #1235: Should stylsheets be embedded or a separate GET?
Right now stylesheets are embedded into the root layout. But, we haven't yet had a very large app. With the boot time of the app being dictated by the the dead render GET then parsing that document and finally making the LV connection if there is a large stylesheet that is embedded into the dead render it will impact the boot time and any re-render of each app.

If we instead link to the stylesheet via a url we could fingerprint the stylesheet and cache it. This way when we re-render the app we can inspect the fingerprint of the url to determine if it should bust the cache or not. If not, we pull from cache rather than GET.

If we go this way I think a larger conversation should be had about a general asset cache. We'd want to cache things like images using the same method. From a simple search it does't seem that `AsyncImage` doesn't cache, but someone shared their solution with a `CachedAsyncImage` view they wrote https://stackoverflow.com/a/69666892/366782

carson-katri:
Closes #1235 

Stylesheet fetching happens concurrently with socket connection. After the socket is connected, we suspend until the stylesheet fetching completes. Multiple `Style` tags will be merged together.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -149,8 +149,17 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
             // extract the root layout, removing anything within the `<div data-phx-main>`.
             let mainDiv = try doc.select("div[data-phx-main]")[0]
             try mainDiv.replaceWith(doc.createElement("phx-main"))
-            self.stylesheet = try? doc.select("Style").reduce(Stylesheet<R>(content: [], classes: [:])) {
-                (try? Stylesheet<R>(from: $1.text(), in: .init()).merge(with: $0)) ?? $0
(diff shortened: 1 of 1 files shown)


bcardarella:
@carson-katri can you cache the result with the url being the key? In dev mode it should always refetch but in prod mode it should fetch from the cache first if it exists

carson-katri:
Caching is built-in to `URLSession`. The user can also change the cache configuration by providing a custom `URLSession` instance to the `LiveSessionConfiguration`.

bcardarella:
@carson-katri yes sorry I recall you mentioning this a few weeks ago now. LGTM!

bcardarella (approved):

bcardarella:
@carson-katri Phoenix support gzip'ing assets as a perf optimization. I believe this will also apply to the LVN stylesheets. Can you confirm if URLSession will pulldown and expand gzip or if that is something you have to account for?

carson-katri:
It's supposed to handle that automatically. I don't think there's a way to stop it from decompressing it either.

PR #1306, opened 2024-04-14
Change styling for utility classes

bcardarella:
This expands the alternate format so now the classes can be written as:

```heex
<Text class ="font(.title) font(.system(size:70)) frame(width:150,height:150)">Foo Bar!</Text>
```

this changes still supports the prior style but this feels a bit more "complete" in that is is a pure 1:1 with the original modifiers, with the exception of whitespace.

I'm considering dropping the prior style in favor of this.

@carson-katri thoughts? I realize you mapped the original class names similar to Tailwind but I'm OK deviating from that. It *does* however start to get confusing as this looks like it should be in a `style` attribute instead.


bcardarella:
The other change was to allow for camel-cased class names rather than split with `-` as the use of the hyphen was ambiguous between the camelcase naming split and the arguments

lib/live_view_native/swiftui/styles/utility.ex
```diff
@@ -119,40 +120,48 @@ defmodule LiveViewNative.SwiftUI.UtilityStyles do
...
     end
   end
 
+  def parse_arguments(<<?(, arguments::binary>>) do
+    <<?), arguments::binary>> = String.reverse(arguments)
+
+    arguments
+    |> String.reverse()
+    |> String.replace(":", ": ")
+    |> String.replace(",", ", ")
```
carson-katri, line 144: Would this mess up string arguments that contain `:` or `,`?

It would be nice if the rules parser didn't require the whitespace and we could send it straight through.

bcardarella: yeah that's a good point

@NduatiK can we support both whitespace and non-whitespacing in the rules parser?

For example:

`foo(bar:1,bar:2)` and `foo(bar: 1, bar: 2)`

NduatiK: Sure thing.

I'm pretty backed up at work so earliest I can get to this is over the weekend.

Issue #1309, opened 2024-04-16
Client doesn't attempt reconnect if first connection fails


bcardarella:
<img width="559" alt="Screenshot 2024-04-16 at 11 50 19 AM" src="https://github.com/liveview-native/liveview-client-swiftui/assets/18524/d7be0e57-1172-4c98-8972-383fe03f7834">

Steps to reproduce:

* new app
* start client and wait for the `Connection Failed` screen
* start server

At this point the client should be attempting periodic reconnects and when the server is available it should recover and reload. This isn't happening.

PR #1311, opened 2024-04-16
Add reconnect menu to `LiveErrorView`

### LINKED ISSUE #1310: Error page can get into unresolvable state
If there is a yet to implemented endpoint and a link is created and followed in the app the app's state is now stuck on that non-existent route. The only solution is to stop and restart the simulator.

It would be good to get a "Restart" button or something to reset the path state manually if this happens.

I ran into this when I by mistake create a `<.link navigate={~p"/charts"}>Chart</.link>`

but the real path is `/chart` not `/charts`. I followed the link in the client and it was now stuck on `/charts` and unrecoverable without a simulator restart.

carson-katri:
Closes #1310

This adds a reconnect button to the `LiveErrorView`. It provides options to reconnect on the current route or from the root.

A new API was also added so users can trigger a reconnect without needing access to the session.
This follows the design of actions like [`openWindow`](https://developer.apple.com/documentation/swiftui/environmentvalues/openwindow) and [`dismiss`](https://developer.apple.com/documentation/swiftui/environmentvalues/dismiss) in SwiftUI.

```swift
struct MyCustomView: View {
  @Environment(\.reconnectLiveView) private var reconnectLiveView

  var body: some View {
    Button("Reconnect") {
      Task {
        await reconnectLiveView()
        await reconnectLiveView(.restart) // use the `restart` style to reconnect from the base URL.
      }
    }
  }
}
```

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/c8f93d37-433e-4cbc-910d-99e725c02cc2

Automatic reconnect is now enabled by default. It can also be customized with the `ReconnectBehavior` type:

```swift
#LiveView(
    .localhost,
    configuration: .init(reconnectBehavior: .exponential(upTo: 10, jitter: 0...1))
)
```

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
@@ -26,6 +26,8 @@ public struct LiveSessionConfiguration {
     /// The transition used when the live view changes connects.
     public var transition: AnyTransition?
     
+    public var reconnectBehavior: ReconnectBehavior = .exponential
+    
     /// Constructs a default, empty configuration.
     public init() {
     }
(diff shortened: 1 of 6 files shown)


bcardarella:
This is perfect 😄

carson-katri:
I'm going to try to get the automatic for #1309 in this PR too.

Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
```diff
@@ -26,17 +26,65 @@ public struct LiveSessionConfiguration {
...
+        private init() {
+            self.delay = nil
+        }
+        
+        /// Attempt to reconnect at an exponential rate.
+        public static func exponential(
+            upTo maxDelay: TimeInterval = 32,
+            step: TimeInterval = 1,
+            jitter: ClosedRange<TimeInterval> = 0...0.25
+        ) -> Self {
```
carson-katri, line 63: @bcardarella Let me know if the defaults here should be tweaked.

bcardarella: I think that's OK for now. We may want to make this configurable in the future.

carson-katri: It already is configurable. The main issue text shows how the API works. Those are just the default values if you don't pass the configuration option.

PR #1324, opened 2024-04-18
Allow empty keyword list as last argument

carson-katri:
Closes https://github.com/liveview-native/live_view_native_stylesheet/issues/62

### DIFF
--- a/Sources/LiveViewNativeStylesheetMacros/ParseableExpressionMacro.swift
+++ b/Sources/LiveViewNativeStylesheetMacros/ParseableExpressionMacro.swift
@@ -189,6 +189,8 @@ public enum ParseableExpressionMacro: ExtensionMacro {
             
             #"try "[".utf8.parse(&input)"#
             
+            #"if input.first != "]".utf8.first {"#
+            
             try WhileStmtSyntax("while !input.isEmpty {") {
                 // switch over the argument name to parse the correct type
                 "try Whitespace().parse(&input)"
@@ -227,6 +229,8 @@ public enum ParseableExpressionMacro: ExtensionMacro {
                 }
                 #"try ",".utf8.parse(&input)"#
             }
+            
+            #"}"#
             
             #"try "]".utf8.parse(&input)"#
             
--- a/Tests/LiveViewNativeStylesheetTests/LiveViewNativeStylesheetTests.swift
+++ b/Tests/LiveViewNativeStylesheetTests/LiveViewNativeStylesheetTests.swift
@@ -23,4 +23,9 @@ final class LiveViewNativeStylesheetTests: XCTestCase {
         XCTAssertEqual(try Alignment.parser(in: .init()).parse("{:., [], [nil, :leading]}"), Alignment.leading)
         XCTAssertEqual(try Alignment.parser(in: .init()).parse("{:., [], [nil, :trailing]}"), Alignment.trailing)
     }
+    
+    func testEvent() throws {
+        XCTAssertNoThrow(try Event.parser(in: .init()).parse(#"{:__event__, [], ["search"]}"#))
+        XCTAssertNoThrow(try Event.parser(in: .init()).parse(#"{:__event__, [], ["search", []]}"#))
+    }
 }


carson-katri:
On second thought, this may need to be fixed on the Elixir side.

If we allow an empty list as the last argument to be an empty keyword list, it could cause ambiguity with a list argument.

```swift
testModifier("arg1")
testModifier("arg1", [])
```

For these modifiers, the first form (with a single string argument) would be preferred even when the empty list is passed. This may not be the correct behavior for that modifier.

```ex
{:testModifier, [], ["arg1", []]}
{:testModifier, [], ["arg1", []]}
```

I think we just need to update the stylesheet to output a consistent AST for the `event` helper.

bcardarella:
I'm still not understanding the background on this.

If I parse the two modifiers currently:

```iex
> LiveViewNative.SwiftUI.RulesParser.parse(~S<testModifier("arg1")>)
{:testModifier, [], ["arg1"]}

> LiveViewNative.SwiftUI.RulesParser.parse(~S<testModifier("arg1", [])>)

LiveViewNative.SwiftUI.RulesParser.parse(~S<testModifier("arg1", [])>)
** (SyntaxError) invalid syntax found on .:1:
   error: Unsupported input:
     |
   1 | testModifier("arg1", [])
     |                      ^^
     |
   
   Expected one of the following:
    - a list of values eg ‘[1, 2, 3]’, ‘["red", "blue"]’ or ‘[Color.red, Color.blue]’
    - a Swift range eg ‘1..<10’ or ‘foo(Foo.bar...Baz.qux)’
    - a number, string, nil, boolean or :atom
    - an event eg ‘event("search-event", throttle: 10_000)’
    - an attribute eg ‘attr("placeholder")’
    - an IME eg ‘Color.red’ or ‘.largeTitle’’
    - a list of keyword pairs eg ‘style: :dashed’, ‘size: 12’ or  ‘style: [lineWidth: 1]’
    - a modifier eg ‘bold()’
    - a variable defined in the class header eg ‘color_name’
``

We can certainly fix the 2nd case so that empty lists are preserved, it would produce:

```elixir
{:testModifier, [], ["arg1", []]}
```

but it seems that you are asking if the first style with a single argument being passed can result in two arguments in the output AST. Is this correct? In other words, you want:

```elixir
> LiveViewNative.SwiftUI.RulesParser.parse(~S<testModifier("arg1")>)
{:testModifier, [], ["arg1", []]}
```

carson-katri:
@bcardarella This is an issued specifically with the `event` helper. The only change we need is for it to encode like this:

```ex
event("name") -> {:__event__, [], ["name"]}
```

Currently, it is adding an empty list to the arguments:

```ex
event("name") -> {:__event__, [], ["name", []]}
```

carson-katri:
Also, there should *not* be two arguments in the output AST if only one is passed in the stylesheet. The current behavior is correct for all modifiers except `event`. If you don't pass any keyword arguments, there should be no `[]` in the argument list.

bcardarella:
Ok now I remember talking about this in the past, it sounded vaguely familiar. I *think* we spoke about it over a call and I cannot find your explanation on why the `[]` cannot represent a zero-arity function call. In Elixir we represent the function in the tuple as `{name, meta, args}` and `args` is always a list. So if there is a zero-arity function it simply has no members in that list of arguments.

carson-katri:
Correct, but there is an empty list in it as well.

`["name", []]`

It should just be `["name"]`

bcardarella:
oh wait, how is the event represented internally? Is the event name itself a function call? I think when this was ideated the concept was that `event("fooBar", 1,2,3)` would call `fooBar(1,2,3)` within the client. Or is the event being called in a differet manner? If the event name isn't a function name then I'm :+1: on changing the Elixir side

carson-katri:
The `event` helper doesn't call anything on the client. It is building a reference to an event on the server.

```elixir
event("fooBar")
```

would call

```elixir
def handle_event("fooBar", _params, socket)
```

The only arguments that are supported are ones that can configure an event, such as `throttle`, `debounce`, `target`, etc.

bcardarella:
thanks that was what I was thinking it would do, just wanted to confirm. I agree this is a bug then and I'll update some tests for @NduatiK to tackle when he is available

carson-katri:
The modifier itself determines what params are sent back to the server.

For example `onHover` passes a bool indicating whether the hover entered or exited.
```swift
onHover(perform: event("hover-changed"))
```
```elixir
def handle_event("hover-changed", true, socket), do: IO.puts("Hover began")
def handle_event("hover-changed", false, socket), do: IO.puts("Hover ended")
```

The `onTapGesture` modifier passes the location of the tap.
```swift
onTapGesture(coordinateSpace: .global, perform: event("tap"))
```
```elixir
def handle_event("tap", tap_location, socket), do: ...
```

bcardarella:
Understood. I'll ad a failing test today for @NduatiK

bcardarella:
Closing in favor of fixing in Elixir

PR #1335, opened 2024-04-30
Add support for visionOS and missing modifiers/types

carson-katri:
This updates the core dependency, with visionOS support included.

The `_GeneratedModifier.swift` was regenerated using Xcode 15.3, and uses the `arm64-apple-xros.swiftinterface` instead of the iOS file to include some missing modifiers. I also added missing types for these modifiers.


bcardarella:
Confirmed that this is working well. There was a MapKit compilation issue but I'll report that in that repo. I think this can move out of Daft

bcardarella (approved):

PR #1337, opened 2024-05-01
Change `_lvn` to `_interface` and send on dead render

### LINKED ISSUE #1321: interface isn't being sent
I am not seeing the interface data that includes the targets (ipados, ipados, macos, tvos, watchos, visionos)

carson-katri:
Closes #1321

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -526,7 +526,6 @@ extension LiveSessionCoordinator {
             "app_version": getAppVersion(),
             "app_build": getAppBuild(),
             "bundle_id": getBundleID(),
-            "format": platform,
             "os": getOSName(),
             "os_version": getOSVersion(),
             "target": getTarget()
@@ -584,25 +583,25 @@ extension LiveSessionCoordinator {
 
     private static func getTarget() -> String {
         #if os(watchOS)
-        return "watchOS"
+        return "watchos"
         #elseif os(macOS)
-        return "macOS"
+        return "macos"
         #elseif os(visionOS)
-        return "visionOS"
+        return "visionos"
         #elseif os(tvOS)
-        return "tvOS"
+        return "tvos"
         #else
         switch UIDevice.current.userInterfaceIdiom {
         case .phone:
-            return "iOS"
+            return "ios"
         case .pad:
-            return "iPadOS"
+            return "ipados"
         case .mac:
-            return "macCatalyst"
+            return "maccatalyst"
         case .tv:
-            return "tvOS"
+            return "tvos"
         case .vision:
-            return "visionOS"
+            return "visionos"
         default:
             return "unknown"
         }
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -304,9 +304,7 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
     /// Returns the dead render HTML and the HTTP response information (including the final URL after redirects).
     func deadRender(for request: URLRequest) async throws -> (String, HTTPURLResponse) {
         var request = request
-        request.url = request.url!.appending(queryItems: [
-            .init(name: "_format", value: "swiftui")
-        ])
+        request.url = request.url!.appendingLiveViewItems(R.self)
         if domValues != nil {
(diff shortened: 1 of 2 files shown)


Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -514,24 +512,21 @@ class LiveSessionURLSessionDelegate<R: RootRegistry>: NSObject, URLSessionTaskDe
...
 extension LiveSessionCoordinator {
     static var platform: String { "swiftui" }
-    static var platformParams: Payload {
+    static var platformParams: [String:String] {
         [
             "app_version": getAppVersion(),
             "app_build": getAppBuild(),
             "bundle_id": getBundleID(),
-            "format": "swiftui",
+            "format": platform,
```
carson-katri, line 529: Do we need to pass the format in the `_interface` too? Or can I remove this.

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -585,22 +584,50 @@ extension LiveSessionCoordinator {
 
     private static func getTarget() -> String {
```
carson-katri, line 585: I updated this to send OS-style names. The only difference between the `target` and `os` arguments is now iPads will return `iPadOS` for the `target`, but `iOS` for the `os`.

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -585,22 +584,50 @@ extension LiveSessionCoordinator {
...
-            return "mac"
+            return "macCatalyst"
         case .tv:
-            return "tv"
+            return "tvOS"
+        case .vision:
+            return "visionOS"
         default:
-            return "unspecified"
+            return "unknown"
```
carson-katri, line 607: What should the default be? It could be changed to `nil`.

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
bcardarella: it can be removed

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
bcardarella: Can you ensure the target names are downcased? I'm up for debating this as we've preserved the upstream casing of the client's UI framework, but in this case within Phoenix the template names and filename segments have always been downcased.

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
bcardarella: Let's go with `"unknown"`

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
carson-katri: Sure, but I think the server may want to check for any casing, even something weird like `app.swiftui+IpAdOs.neex`. Not all filesystems are case sensitive like APFS though.

bcardarella: The issue isn't the filesystem, the issue is Elixir's own function file finding files by glob: https://hexdocs.pm/elixir/Path.html#wildcard/2 is case-sensitive. It's used in `Phoenix.Template` https://github.com/phoenixframework/phoenix_template/blob/main/lib/phoenix/template.ex#L352 for finding the relevant template and I'm leaning into that implementation

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -585,22 +583,50 @@ extension LiveSessionCoordinator {
...
         switch UIDevice.current.userInterfaceIdiom {
         case .phone:
-            return "phone"
+            return "ios"
         case .pad:
-            return "pad"
+            return "ipados"
         case .mac:
-            return "mac"
+            return "maccatalyst"
```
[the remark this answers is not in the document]
bcardarella: oh wait, you cover that above. What circumstance would this be reached if LOC 588 is covering `macos` ?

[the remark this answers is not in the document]
carson-katri: It would be the `macOS (Designed for iPad)` option afaik.

PR #1338, opened 2024-05-01
Disable change tracked `local` form state mode

carson-katri:
Now, any element that uses `@FormState` won't send `phx-change` events unless it is nested within a `<LiveForm>`. It can be interacted with, but the entered values will not be persisted when outside of a form.

```heex
<TextField phx-change="changed">Name</TextField>
```
This will log a warning as well:
```
Form element used outside of a <LiveForm>. This may not behave as expected.
```

When inside a `LiveForm`, change events will be sent as normal:
```heex
<LiveForm id="test">
  <TextField name="text-field" phx-change="changed">Name</TextField>
</LiveForm>
```
```
[debug] HANDLE EVENT "changed" in HelloWorldWeb.HomeLive
  Parameters: %{"text-field" => "Testing"}
```

Related PR: https://github.com/liveview-native/liveview-native-live-form/pull/14

### DIFF
--- a/Sources/LiveViewNative/Property Wrappers/FormState.swift
+++ b/Sources/LiveViewNative/Property Wrappers/FormState.swift
@@ -134,7 +134,6 @@ public struct FormState<Value: FormValue> {
         }
         nonmutating set {
             resolveMode()
-            let oldValue = wrappedValue
             switch data.mode {
             case .unknown:
                 fatalError("@FormState cannot be accessed before being installed in a view")
@@ -145,11 +144,11 @@ public struct FormState<Value: FormValue> {
                     logger.log(level: .error, "Expected @FormState in form mode to have element with name")
                     return
                 }
-                formModel[elementName] = newValue
-                // todo: this will send a change event for both the form and the input if the input has one, check if that matches web
-            }
-            if oldValue != newValue {
-                sendChangeEventIfNecessary()
+                formModel.setValue(
+                    newValue,
+                    forName: elementName,
+                    changeEvent: sendChangeEvents ? element.attributeValue(for: "phx-change") : nil
+                )
             }
         }
     }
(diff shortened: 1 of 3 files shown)


bcardarella:
Can you add a documentation note to each of the generated hex docs for each of these views that `phx-change` is only supported when the element is wrapped with `<LiveForm>` ?

I'll check with @chrismccord but I think that there is also a requirement that the parent `<form>` must have a `phx-change` event for the child input's `phx-change` to be active. [Source](https://hexdocs.pm/phoenix_live_view/form-bindings.html)

> You may wish for an individual input to use its own change event or to target a different component. This can be accomplished by annotating the input itself with phx-change, for example:

```elixir
<.form for={@form} phx-change="validate" phx-submit="save">
  ...
  <.input field={@form[:email]}  phx-change="email_changed" phx-target={@myself} />
</.form>
```

carson-katri:
I tested this, and `phx-change` was not required on the `<.form>` for it to work on an `<.input>`.

However, it did send some extra information:

```ex
%{"_target" => ["text-field"], "text-field" => ""}
```

We aren't sending that `_target` value. I assume it would just match the `name` of the `<.input>`.

carson-katri:
I found a few more features of form that we don't support:

1. [Recovery from crashes or disconnects](https://hexdocs.pm/phoenix_live_view/form-bindings.html#recovery-following-crashes-or-disconnects)
2. [Resetting Forms](https://hexdocs.pm/phoenix_live_view/form-bindings.html#resetting-forms)

bcardarella:
I'll try to get more information on this today.

For the two other features, let's discuss on standup

carson-katri:
Looks like `phx-change` on the `<.form>` only sends the element that changed, just like if you had put it directly on the `<.input>`. It also seems like a `phx-change` on the `<.input>` will prevent the `phx-change` on the `<.form>` from triggering.

We might need to adjust how this is being handled in the client and live form libraries.

bcardarella:
@carson-katri for the form that sent `%{"_target" => ["text-field"], "text-field" => ""}` what did the form markup look like?

bcardarella:
It looks like Phoenix is using the `_target` for form reset: https://hexdocs.pm/phoenix_live_view/form-bindings.html#resetting-forms

What is confusing is why the `text-field` key wasn't nested under the form's changeset schema.

bcardarella:
@carson-katri what is the value of `@form` ?

carson-katri:
Just a map in the assigns:

```
{:ok, assign(socket, form: %{ "value" => "test" })}
```

bcardarella:
is this good to merge?

PR #1352, opened 2024-05-09
Support `phx-debounce="blur"`

### LINKED ISSUE #1322: phx-debounce doesn't support blur
LV docs: https://hexdocs.pm/phoenix_live_view/bindings.html#rate-limiting-events-with-debounce-and-throttle

`phx-debounce` should accept both the ms time but also `blur` as a value. So the change event shouldn't trigger until the input itself is unfocused.

carson-katri:
Closes #1322 

`phx-debounce="blur"` only works on `<TextField>`, `<SecureField>`, and `<TextEditor>`. Other controls (like `<Slider>` or `<Toggle>`) don't seem capable of receiving focus. However, it could easily be expanded to other form controls as needed.

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/b6415b76-88c9-46ab-9bd4-66ff35ef615e


bcardarella:
@carson-katri can you resolve these conflicts for merging?

PR #1359, opened 2024-05-15
Failing test for parser update

### LINKED ISSUE #1358: `event(...)` without keyword arguments has inconsistent encoding
Typically, when a modifier has no keyword arguments, it is encoded like so:
```elixir
my_event("search") -> {:my_event, [], ["search"]}
```

However, when `event` is used, it is encoded with an empty list at the end:

```elixir
event("search") -> {:__event__, [], ["search", []]}
```

This keyword list should only be present when a labelled argument is passed to `event`:

```elixir
event("search", debounce: 100) -> {:__event__, [], ["search", [debounce: 100]]}
```

bcardarella:
Closes #1358


NduatiK:
@bcardarella, fixed

PR #1366, opened 2024-05-22
Use `style` attribute for generated docs

carson-katri:
This will automatically switch between using double `"` and single `'` quotes depending on the contents of the modifier being documented.

I've also updated the generator to use the XROS swiftinterface file instead of the iPhoneOS one to include visionOS modifiers in the generated documentation.

Screenshot 2024-05-22 at 4 10 26 PM


bcardarella:
I don't think the `<%!-- template -- %>` parts are necessary

PR #1368, opened 2024-05-23
Expose `node` and `data` on ElementNode

carson-katri:
This opens up the `ElementNode` API for addon libraries.

They can now access the `Node` from an `ElementNode` instance. I've also added an `Identifiable` conformance to `ElementNode`, where the ID of the element node is the `node.id`


bcardarella:
What is this for again?

carson-katri:
Nothing specific yet, just a general extension of the API for addon authors.

PR #1369, opened 2024-05-23
Update core components to use `style` attribute

carson-katri:
This replaces the use of utility classes with inline styles in the generate `core_components.swiftui.ex` file.
All components should behave the same.

### DIFF
--- a/priv/templates/lvn.swiftui.gen/core_components.ex
+++ b/priv/templates/lvn.swiftui.gen/core_components.ex
@@ -44,7 +44,7 @@ defmodule <%= inspect context.web_module %>.CoreComponents.<%= inspect context.m
 
   ## Examples
 
-      <Group class="keyboardType(.emailAddress)">
+      <Group style="keyboardType(.emailAddress)">
         <.input field={@form[:email]} type="TextField" />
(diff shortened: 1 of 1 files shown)


priv/templates/lvn.swiftui.gen/core_components.ex
```diff
@@ -251,18 +251,17 @@ defmodule <%= inspect context.web_module %>.CoreComponents.<%= inspect context.m
   """
   @doc type: :component
 
-  attr :class, :string, default: nil
```
bcardarella, line 254: I think we'll want to retail the `class` attr on these components. We just need to add `style`

carson-katri: This component just adds content to the navigation bar (for the title, subtitle, or toolbar). So adding additional classes shouldn't have any effect.

carson-katri: In other cases, `class` should be caught by `:rest`.

bcardarella: yes, good point

PR #1378, opened 2024-06-19
2024 platform updates support

carson-katri:
This adds support in the `ModifierGenerator` for new modifiers in iOS 18 and aligned releases.

SwiftUI was split into `SwiftUICore` and `SwiftUI` frameworks. The modifier generator now accepts an array of `.swiftinterface` files to support this restructuring.

Here are some highlights:

1. `navigationTransition(_:)`

This modifier can be used to perform a `zoom` transition when navigating, instead of the typical push animation.

https://github.com/liveview-native/liveview-client-swiftui/assets/13581484/b254dd5c-1990-4ecd-9341-f40ab3a0537d

> [!NOTE]
> The animation is not yet correct on backward navigation. This is because the previous page is reloaded when navigating back, causing the source element to be rebuilt with a new identity during the animation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,8 +52,8 @@ jobs:
       - name: Build for watchOS
         shell: bash
         run: |
-          xcodebuild -scheme LiveViewNative-Package -destination "platform=watchOS Simulator,name=Apple Watch Series 9 (45mm)" -skipMacroValidation -skipPackagePluginValidation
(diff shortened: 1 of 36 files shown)


bcardarella:
@carson-katri do we let this wait until ios18 is released or can this be merged in now?

carson-katri:
This should wait until Xcode 16 is released. The new APIs are also likely to change between betas, so we may need to regenerate the modifiers as betas are released and test again.

bcardarella:
Ok, I'll  create a xcode-16 tag then for all of these things

PR #1389, opened 2024-07-18
Automatic scroll restoration for `ScrollView`

carson-katri:
Scroll restoration is now automatically provided for all `ScrollView` elements with an explicit `id` attribute.

```html
<ScrollView id="cookbook">
  ...
</ScrollView>
```

Scroll restoration only occurs on back navigation. It is also only available for the `ScrollView` element (not `List` or other scrolling containers)

> [!NOTE]
> Scroll restoration is only available for iOS 18 or later.

https://github.com/user-attachments/assets/dce871ae-d628-4145-9ab3-492971a087ae


bcardarella:
Can you update the CHANGELOG? This should probably go under `Fixed`

bcardarella (approved):

PR #1406, opened 2024-08-08
Add access to `LiveContext` from `ContentBuilder`

carson-katri:
This fixes instances in addon libraries where a blank (unusable) `LiveContext<R>()` had to be passed to resolve a stylesheet value.

https://github.com/liveview-native/liveview-native-swiftui-charts/blob/c612c5ac085fc61a33a4c3d097d09b0db9a2be7f/Sources/LiveViewNativeCharts/Modifiers/SymbolModifier.swift#L44

Instead, a valid `LiveContext` can be accessed from the `ContentBuilder` context, via `context.context`.

Addon libraries will need to be updated to make use of this new API.

### DIFF
--- a/Sources/LiveViewNative/Property Wrappers/Event.swift
+++ b/Sources/LiveViewNative/Property Wrappers/Event.swift
@@ -144,21 +144,21 @@ public struct Event: DynamicProperty, Decodable {
             if let debounce = debounce {
                 handlerTask = Task { [weak didSendSubject] in
                     for await event in channel.debounce(for: .milliseconds(debounce)) {
-                        try await coordinator?.pushEvent(event.0, event.1, event.2, event.3)
+                        _ = try await coordinator?.pushEvent(event.0, event.1, event.2, event.3)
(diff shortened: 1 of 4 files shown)


carson-katri:
Here's one example of what this fixes:

Without a valid context, a `Chart` with a modifier that embeds a View (such as this `Image` rendering for each `PointMark`) would crash. Now we can render these templates properly:

```html
<Chart style="aspectRatio(1.5, contentMode: .fit); foregroundStyle(.orange); padding();">
  <PointMark
    :for={i <- 1..10}
    x:label="X"
    x:value={i}
    y:label="Y"
    y:value={:rand.uniform()}

    style="symbol(symbol: :symbol);"
  >
    <Image systemName="bolt" template="symbol" style={[
      "font(.title3)",
      "symbolVariant(.circle.fill)",
      "symbolRenderingMode(.hierarchical)"
    ]} />
  </PointMark>
</Chart>
```

Screenshot 2024-08-08 at 3 15 56 PM

PR #1417, opened 2024-08-23
Fix document reference

dbii:
Change the link from Installation Guide to Getting Started and point to the current URL as the old one no longer exists.


BrooklinJazz (approved):
Looks great! @bcardarella can we merge?

PR #1424, opened 2024-08-29
Fix generated errorview tvos

dbii:
Menu is only available on tvOS 17+. It is used in the generated Swift ErrorView.swift.


Sources/LiveViewNative/Live/LiveErrorView.swift
```diff
@@ -39,25 +39,27 @@ public struct LiveErrorView<Fallback: View>: View {
                 }
                 .padding()
                 #else
-                SwiftUI.Menu {
-                    SwiftUI.Button {
-                        Task {
-                            await reconnectLiveView(.automatic)
+                if #available(iOS 14, macOS 11, tvOS 17, visionOS 1, *) {
```
carson-katri, line 42: This should probably have an `else` branch to fallback on a regular `Button` to reconnect from root (like on watchOS).

dbii: I added an `else` with a button. I think I did it correct, but it should be reviewed.

carson-katri (approved):

PR #1426, opened 2024-08-30
Support version normalizations

bcardarella:
Depends on: https://github.com/liveview-native/live_view_native/pull/200


bcardarella:
The failing CI is on the Swift side which this PR doesn't touch. We'll have to investigate later what's happening there. I tried restarting CI and no luck.

PR #1435, opened 2024-09-03
Navigation optimizations

carson-katri:


Sources/BuiltinRegistryGenerator/BuiltinRegistryGenerator.swift
```diff
@@ -43,7 +44,7 @@ struct BuiltinRegistryGenerator: ParsableCommand {
         "RoundedRectangle": "Shape<R, RoundedRectangle>(shape: RoundedRectangle(from: element))",
         "Color": "ColorView<R>()",
         "Image": "ImageView<R>()",
-        "phx-main": "PhxMain<R>()"
+        "div": "PhxMain<R>()"
```
carson-katri, line 47: @bcardarella I had to use some workarounds to get SwiftSoup removed. This is one of them, we don't have a way to replace a tag in a `Document` from the client in liveview-native-core. To get this merged for v0.3.1, I'll need to revert back to using SwiftSoup.

Although I feel like LiveView Native shouldn't send a `div` element for the LiveView entrypoint anyways.

carson-katri: This has been reverted

bcardarella (approved):

Issue #1436, opened 2024-09-05
Wrong attribute case for system images in SwiftUI Views guide


egutter:
https://hexdocs.pm/live_view_native_swiftui/swiftui-views.html#system-images

Replace system image attributes using kebab case `system-image` or `system-name` to camel case: `systemImage` or `systemName`

bcardarella:
Hi did you want to PR this fix?

bcardarella:
Closed by #1437

PR #1437, opened 2024-09-05
Replace system image attributes with kebab case to camel case.

egutter:
Fix issue https://github.com/liveview-native/liveview-client-swiftui/issues/1436


bcardarella:
Thank you!

PR #1439, opened 2024-09-10
Core networking

carson-katri:


Sources/LiveViewNative/__CoreExtensions.swift
simlay, line 1: @mobile-bungalow if I understand correctly, these changes should be applied to [`Support.swift` in core](https://github.com/liveview-native/liveview-native-core/blob/49ee4643d9c10c0f16ec7432c3e636832c0eed4b/crates/core/liveview-native-core-swift/Sources/LiveViewNativeCore/Support.swift).

PR #1440, opened 2024-09-10
Improve structure of generated docs

carson-katri:


bcardarella:
This is great, I was working on something similar but got stalled out prior to ElixirConf so thanks for tackling this!

PR #1446, opened 2024-09-17
Apply various @MainActor annotations

nighthawk:
In Xcode 16 `View` automatically gets `@MainActor` applied. This breaks creating a custom LVN Addon as the `static lookup(_:,element:)` is not isolated to the main actor, which the Swift compiler doesn't allow. 

I went through the exercise of annotating `CustomRegistry` as `@MainActor` and working through the many follow-ups of annotating other things also as `@MainActor`. 

This is now working in my early test cases, but I'm by no means sure that this is the right approach here. Hence, marking this as a draft for now.

### DIFF
--- a/Sources/LiveViewNative/_GeneratedModifiers.swift
+++ b/Sources/LiveViewNative/_GeneratedModifiers.swift
@@ -13562,6 +13562,7 @@ case let ._MatchedGeometryEffectModifier(modifier):
             
             let context: ParseableModifierContext
             
+            @MainActor
             func parse(_ input: inout Substring.UTF8View) throws -> Output {
                 let parsers = [
                     _accessibilityActionModifier<R>.name: _accessibilityActionModifier<R>.parser(in: context).map({ Output.chunk0(.accessibilityAction($0)) }).eraseToAnyParser(),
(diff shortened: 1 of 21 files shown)


Sources/LiveViewNative/_GeneratedModifiers.swift
```diff
@@ -13562,6 +13562,7 @@ case let ._MatchedGeometryEffectModifier(modifier):
             
             let context: ParseableModifierContext
```
nighthawk, line 13564: This file seems auto-generated, so this fix isn't enough as it'd get reverted on the next generation.

carson-katri:
Can you share an example of what the code that causes the build issue? I'm assuming you built with the `Swift Language Version` set to `Swift 6`

carson-katri:
In the meantime, you may be able to import `LiveViewNative` differently to work around this:

```
@preconcurrency import LiveViewNative
```

nighthawk:
I indeed had `swift-tools-version: 6.0` in my `Package.swift` which turned on the strict concurrency checks. I've reverted that back to `5.x` and that's working fine with version `0.3.0` of LiveView Native.

There wasn't anything special in my package, so the same would happen in other add-on libraries if they switch to the 6.0 language mode.

For reference, this is my add-on package: [LiveViewNativeGeoProjector](https://github.com/maparoni/liveview-native-swiftui-geoprojector/tree/main/Sources/LiveViewNativeGeoProjector)

nighthawk:
Closing this due to the many merge conflicts.

PR #1449, opened 2024-09-17
Query encode `+` characters in form events

### LINKED ISSUE #1447: Form fields replace "+" with a space when sent as params
I noticed this when trying to create a new user account using `+` in an email:

Screenshot 2024-07-24 at 5 04 01 PM

Server log with params received:

```elixir
[debug] HANDLE EVENT "validate" in LaxWeb.UserRegistrationLive
  Parameters: %{"_target" => ["user", "email"], "user" => %{"email" => "email trailing@gmail.com"}}
```

carson-katri:
Closes #1447

### DIFF
--- a/Sources/LiveViewNative/ViewModel.swift
+++ b/Sources/LiveViewNative/ViewModel.swift
@@ -93,7 +93,7 @@ public class FormModel: ObservableObject, CustomDebugStringConvertible {
     
     /// Create a URL encoded body from the data in the form.
     public func buildFormQuery() throws -> String {
-        return try buildFormURLComponents().query!
+        return try buildFormURLComponents().formEncodedQuery!
     }
     
     private func sendChangeEventForFormElement(_ value: any FormValue, for name: String, _ sendEvent: ((Any) async throws -> ())?) -> (() async throws -> Void)? {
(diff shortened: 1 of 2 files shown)


Sources/LiveViewNative/ViewModel.swift
```diff
@@ -213,6 +213,6 @@ private extension URLComponents {
         else { return nil }
         // `+` is converted to a space, so query encode it.
         // Spaces are already converted to %20 by `percentEncodedQuery`
-        return query.replacingOccurrences(of: "+", with: "%2B")
+        return query.replacing("+", with: "%2B")
```
bcardarella, line 216: is there a more comprehensive encoding rather than a single one off?

carson-katri: I'm using the built-in percent encoding here: https://github.com/liveview-native/liveview-client-swiftui/pull/1449/files#diff-8a9b2e97113616f129b2f083e64768aeb40d0033ee8d7f20cc42f7940c63981bR212

But it doesn't cover the `+` character, because this is technically valid in query params. Another option would be to do a fully query encode of each value, only accepting a small character set (such as `.alphanumeric`) instead of `urlQueryAllowed`.

bcardarella:
@carson-katri we're getting failing CI

PR #1455, opened 2024-09-19
Fix retain cycle in LiveViewCoordinator

nighthawk:
`LiveViewCoordinator` is initialised from `LiveSessionCoordinator`, which gets passed in as a parameter to `init` and then retained by the view coordinator. This causes a retain cycle.

Additionally, this is also retained by the `handleEvent` as `session` was used directly in the closure, which then retained it.

This commit fixes both by using a weak reference from the view coordinator to the session coordinator, and accessing the session via the weak `self` reference.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
@@ -32,7 +32,7 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
         internalState
     }
     
-    @_spi(LiveForm) public weak var session: LiveSessionCoordinator<R>!
+    @_spi(LiveForm) public private(set) weak var session: LiveSessionCoordinator<R>!
     var url: URL
     
     private var channel: Channel?
--- a/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
@@ -32,7 +32,7 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
         internalState
     }
     
-    @_spi(LiveForm) public let session: LiveSessionCoordinator<R>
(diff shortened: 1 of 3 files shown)


nighthawk:
The retain cycle is triggered by `LiveSessionCoordinator`'s `init` method which has this line:

```swift
self.navigationPath = [.init(url: url, coordinator: .init(session: self, url: self.url))]
```

Note how `self` is passed there, which causes the retain cycle.

I noticed this when I didn't have my local server running and the LiveView in the app couldn't connect, and when I navigated away from the LiveView in my app and it was no longer in the view hierarchy, the live view session kept trying to reconnect.

nighthawk:
Seems like that retain cycle hid two other issues here. Since `LiveSessionCoordinator` will now get deallocated properly, the two `unused self` usages need to be removed as they can fire in rare cases after it got deallocated, which then crashes.

carson-katri (changes requested):
Great catch, thanks!

Could you update the CHANGELOG.md file to note this fix?

bcardarella:
This should go under `Fixed` in the `CHANGELOG.md`

Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
```diff
@@ -32,7 +32,7 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
         internalState
     }
     
-    @_spi(LiveForm) public let session: LiveSessionCoordinator<R>
+    @_spi(LiveForm) public weak var session: LiveSessionCoordinator<R>!
```
carson-katri, line 35: This should be `private(set)`
```suggestion
    @_spi(LiveForm) public private(set) weak var session: LiveSessionCoordinator<R>!
```

nighthawk:
Sure thing. Change log updated.

carson-katri (approved):
LGTM 👍

bcardarella:
@nighthawk thank you!

PR #1456, opened 2024-09-19
Allow to set optional headers

nighthawk:
These are passed on to Phoenix when the LiveView makes the initial connection. 

Sample use case: Passing an "Authorization" HTTP header to authenticate a user in the LiveView with a token that's fetched outside of LiveView and stored in the user's keychain. That particular header should not be set on [`URLSessionConfiguration`](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1411532-httpadditionalheaders) directly.

[Related Slack thread](https://elixir-lang.slack.com/archives/C041F3EQHFS/p1726716326675869) for reference.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
@@ -15,6 +15,11 @@ public struct LiveSessionConfiguration {
     ///
     /// By default, no connection params are provided.
     public var connectParams: ((URL) -> [String: Any])? = nil
+  
+    /// Optional headers that are passed along with estabilishing the initial connection.
+    ///
+    /// By default, no addtional headers are provided.
+    public var headers: [String: String] = [:]
     
     /// The URL session configuration the coordinator will use for performing HTTP and socket requests.
     /// 
@@ -34,10 +39,12 @@ public struct LiveSessionConfiguration {
     
     public init(
         connectParams: ((URL) -> [String: Any])? = nil,
+        headers: [String: String] = [:],
         urlSessionConfiguration: URLSessionConfiguration = .default,
         transition: AnyTransition? = nil,
         reconnectBehavior: ReconnectBehavior = .exponential
     ) {
+        self.headers = headers
         self.connectParams = connectParams
         self.urlSessionConfiguration = urlSessionConfiguration
         self.transition = transition
--- a/Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
@@ -16,7 +16,7 @@ public struct LiveSessionConfiguration {
     /// By default, no connection params are provided.
     public var connectParams: ((URL) -> [String: Any])? = nil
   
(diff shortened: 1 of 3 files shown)


Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -326,6 +326,9 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
         if domValues != nil {
             request.setValue(domValues.phxCSRFToken, forHTTPHeaderField: "x-csrf-token")
         }
+        for (header, value) in configuration.headers {
+            request.setValue(value, forHTTPHeaderField: header)
+        }
```
carson-katri, line 331: Can we change the order these are applied?

```suggestion
        request.allHTTPHeaderFields = configuration.headers
        
        if domValues != nil {
            request.setValue(domValues.phxCSRFToken, forHTTPHeaderField: "x-csrf-token")
        }
```

carson-katri:
@simlay We'll need a way to pass headers to the dead render in core for v0.4.0.

Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
```diff
@@ -15,6 +15,11 @@ public struct LiveSessionConfiguration {
     ///
     /// By default, no connection params are provided.
     public var connectParams: ((URL) -> [String: Any])? = nil
+  
+    /// Optional headers that are passed along with estabilishing the initial connection.
+    ///
+    /// By default, no addtional headers are provided.
+    public var headers: [String: String] = [:]
```
carson-katri, line 22: Can we make this optional with a default value of `nil`? This would match the `URLRequest` API, and makes it more clear that the headers are optional.

Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
```diff
@@ -15,6 +15,11 @@ public struct LiveSessionConfiguration {
     ///
     /// By default, no connection params are provided.
     public var connectParams: ((URL) -> [String: Any])? = nil
+  
+    /// Optional headers that are passed along with estabilishing the initial connection.
```
carson-katri, line 19:
```suggestion
    /// Optional headers that are sent when fetching the dead render.
```

carson-katri:
Can you add a line to the `CHANGELOG.md` file mentioning the new parameter?

bcardarella:
@nighthawk this should go under `Added` in the `CHANGELOG.md`

Sources/LiveViewNative/Coordinators/LiveSessionConfiguration.swift
nighthawk: Done.

nighthawk:
@carson-katri Code review addressed and updated the change log, too.

bcardarella:
thank you, @carson-katri will be able to sign off on this on Tuesday

carson-katri (approved):
LGTM 👍

bcardarella:
Thank you!

PR #1458, opened 2024-09-21
Prepare SwiftUI parser for outputing stylesheet as a map

### LINKED ISSUE #1457: Change stylesheet output to JSON
We currently emit Elixir but I believe for various reasons it will be better to emit JSON.

1. JSON parsing is available in nearly every language so future clients won't have to implement a new parser, Elixir language parsing is not or has incomplete parsing
2. JSON parsing is very fast in most languages
3. We are already encoding some attribute values in JSON

I don't believe we need to change how we are currently constructing our AST, so we will keep that in Elixir format but before it writes to the stylesheet it will convert to JSON

NduatiK:
Prepare for liveview-native/live_view_native_stylesheet#86

Prepare ast for conversion to JSON
- Remove variable related code since we are using EEx
- Always return a list of modifiers (instead of the inner tuple for singletons)
- Return annotations as a map instead of keyword list (since empty keywords are treated as lists)


bcardarella:
I think there was a misunderstanding. We don't need the individual rules output changing at all. Infact I don't think we need any actual changes to how we're generating the stylesheet currently. Instead in `live_view_native_stylesheet` we simply need to convert the currently generated map to the json object. We can then use https://erlang.org/documentation/doc-15.0-rc3/lib/stdlib-6.0/doc/html/json.html#encode/2 to convert values like tuples that are not generically supported

NduatiK:
Aah, my bad I wasn't very clear when asking Question 2. I've updated the SwiftUI parser to generate an AST match the spec you shared (changes in first comment).

Here's the functionality we need to convert the AST into JSON, but I'm not sure where it should be called. Where do you build the final style asset?

```elixir
def ast_to_json(ast) do
  ast
  |> :json.encode(&ast_encoder/2)
  |> IO.iodata_to_binary()
end

defp ast_encoder(nil, encode) do
  :json.encode_value(:null, encode)
end

defp ast_encoder(value, encode) when is_tuple(value) do
  :json.encode_list(Tuple.to_list(value), encode)
end

defp ast_encoder([{_, _} | _] = value, encode) do
  :json.encode_key_value_list(value, encode)
end

defp ast_encoder(other, encode) do
  :json.encode_value(other, encode)
end
```

bcardarella:
I believe this is invalidated by https://github.com/liveview-native/live_view_native_stylesheet/pull/89. The internal representation shouldn't need to change as we are handling the proper render encoding in the json encoder

@NduatiK close?

PR #1459, opened 2024-09-23
Adjust examples in documentation

Francisco-Castro:
This PR adjusts two examples in the documentation.

### DIFF
--- a/livebooks/markdown/stylesheets.md
+++ b/livebooks/markdown/stylesheets.md
@@ -464,7 +464,7 @@ Then use the `template` attribute on the view to be injected into the stylesheet
 
 ```heex
 <Image style="overlay-circle">
-  <Circle template="circle" style="stroke(.red, lineWidth: 4)" >
+  <Circle template="circle" style="stroke(.red, lineWidth: 4)"></Circle>
 </Image>
 ```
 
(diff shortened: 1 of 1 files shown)


livebooks/markdown/stylesheets.md
```diff
@@ -464,7 +464,7 @@ Then use the `template` attribute on the view to be injected into the stylesheet
 
 ```heex
 <Image style="overlay-circle">
-  <Circle template="circle" style="stroke(.red, lineWidth: 4)" >
+  <Circle template="circle" style="stroke(.red, lineWidth: 4)"></Circle>
```
bcardarella, line 467: This can just be self closing:

```heex
<Cirble template="circle" style="stroke(.red, lineWidth: 4)" />
```

Francisco-Castro: Oh, I see! It’s missing the backslash. By the way, how do you know when can be a self-closing element?

@bcardarella I just pushed a commit with the updated version.

bcardarella:
thank you!

livebooks/markdown/stylesheets.md
carson-katri: I believe any element can be self-closing if it doesn't necessitate any child content.

bcardarella: @Francisco-Castro as @carson-katri said, I have been opting into self-closing when there are no children present as it just looks cleaner in the markup but you are free to use a closing tag if you want in your own apps!

Francisco-Castro: That makes sense. I like that approach of opting for self-closing tags when there are no more children present—it definitely makes the markup look cleaner. Thank you both for clarifying!

Issue #1460, opened 2024-09-24
[Bug]: `StylesheetCache` is not thread-safe


nighthawk:
### What happened?

I'm experiencing occasional runtime crashes from `StylesheetCache`. 

It's accessed from `LiveSessionCoordinator.connect()` like this:

```swift
            async let stylesheet = withThrowingTaskGroup(of: Stylesheet<R>.self) { group in
                for style in try doc.select("Style") {
                    guard let url = URL(string: try style.attr("url"), relativeTo: url)
                    else { continue }
                    group.addTask {
                        if let cachedStylesheet = StylesheetCache[for: url, registry: R.self] {
                            return cachedStylesheet
                        } else {
                            let (data, response) = try await self.urlSession.data(from: url)
                            guard let contents = String(data: data, encoding: .utf8)
                            else { return Stylesheet<R>(content: [], classes: [:]) }
                            let stylesheet = try Stylesheet<R>(from: contents, in: .init())
                            StylesheetCache[for: url, registry: R.self] = stylesheet // <= 💥
                            return stylesheet
                        }
                    }
                }
                return try await group.reduce(Stylesheet<R>(content: [], classes: [:])) { result, next in
                    return result.merge(with: next)
                }
            }
```

This then crashes as `StylesheetCache` is not thread-safe:

<img width="1544" alt="image" src="https://github.com/user-attachments/assets/c6878ec0-f829-4c15-a961-8ae290dc6238">

It's probably better to implement this as an `actor` or similar.

I'm running in Swift 5.0 language mode, i.e., without the strict 6.0 concurrency.

### Library Version

0.3.0

### Xcode Version

16.0

### Swift Version

6.0

### On which device or simulator are you running into the problem?

iPad

### Target Device Operating System Version

18.0

### Relevant log output

```shell
[Socket] SwiftPhoenixClient: Socket attempting to reconnect
```

PR #1461, opened 2024-09-24
Turn `StylesheetCache` into a thread-safe actor

nighthawk:
Fixes the issue raised in #1460.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updates to change-tracked properties no longer occur on the next RunLoop, fixing modal dismissal on macOS (#1450)
 - `+` characters are properly encoded as `%2B` in form events (#1449)
 - Fixed retain cycle in `LiveViewCoordinator` (#1455)
+- Made `StylesheetCache` thread-safe, fixing occasional crashes (#1461)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updates to change-tracked properties no longer occur on the next RunLoop, fixing modal dismissal on macOS (#1450)
 - `+` characters are properly encoded as `%2B` in form events (#1449)
 - Fixed retain cycle in `LiveViewCoordinator` (#1455)
+- Made `StylesheetCache` thread-safe, fixing occasional crashes (#1461)
```
carson-katri, line 23: I'm not sure we need to note this in the changelog, since the `StylesheetCache` is not part of any release yet.

carson-katri (commented):
I like this refactor, thanks!

CHANGELOG.md
bcardarella: if the stylesheet cache was crashing the app due to a thread safety issue I'm OK acknowledging that here

carson-katri: We have no note that `StylesheetCache` was added, so we should probably add that to the changelog too then.

bcardarella: I think the cache was added before the CHANGELOG was added though, or am I mistaken?

PR #1464, opened 2024-10-02
Use value from document as default for unbound `FormState`

### LINKED ISSUE #1438: [Bug]: Toggle isOn not working
### What happened?

The toggle is not showing the correct state.

```
  <Toggle isOn={true} >
  Should be on
  </Toggle>
```
IMG_203F238F319A-1

### Library Version

0.3.0

### Xcode Version

lvn go

### Swift Version

lvn go

iPhone

17.6.1

### Relevant log output

_No response_

carson-katri:
Closes #1438 

Previously, `FormState` would use the client-provided fallback value instead of the value in the document when not bound to a `LiveForm`.

Now, any values passed in the template will be used. `Toggle` will reflect the `isOn` attribute, `TextField` will show the provided `text` content, etc. Any fields not bound to a form with a default value cannot be updated. They will revert back to the state provided by the template if modified.

https://github.com/user-attachments/assets/3e1dc603-35ac-40af-b227-c097d471054f

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `+` characters are properly encoded as `%2B` in form events (#1449)
 - Fixed retain cycle in `LiveViewCoordinator` (#1455)
 - Made `StylesheetCache` thread-safe, fixing occasional crashes (#1461)
+- Form elements outside of a `LiveForm` will show the value provided in the template (#1464)
 
 ## [0.3.0] 2024-08-21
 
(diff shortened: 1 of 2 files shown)


bcardarella:
Why does it revert? I don't think we'd want that as it doesn't match the behavior of the web.

If the state of the value bound is updated server-side does the client update too despite the view not being within a `LiveForm`?

carson-katri:
It's because it doesn't have any local state backing it, just a static value from the server. When the server changes the value in the attribute, the field will update.

I can change it so you can update the field, and it will only "revert" when the server changes the value to something else.

carson-katri:
The behavior has been updated to match the web.

PR #1472, opened 2024-10-09
Use the dead render for an initial render and merge into it

carson-katri:
Here you can see the dead render as a red screen and the connected render as a green screen:

https://github.com/user-attachments/assets/f0fd0dba-05a7-43d6-b839-b9ae39b4b5e9


bcardarella:
Do you know if this would allow us to change the navigation type in root?

carson-katri:
Do you mean a separate navigation structure depending on the route?

bcardarella:
Yes, the dead render will only happen if a new connection is established so can we use that as a way to allow for swapping out the navigation type?

carson-katri:
You could use a `live_session` with separate root layouts. I don't think anything prevented that before though.

bcardarella:
Ready to merge?

carson-katri:
If we're doing core integration for v0.4.0, I think this should be included as part of #1439

PR #1485, opened 2024-11-13
Add action to check if changelog was updated

carson-katri:


.github/workflows/changelog_updated.yml
```diff
@@ -15,4 +15,4 @@ jobs:
         with:
           fetch-depth: 2
       - name: Check if CHANGELOG.md was changed
-        run: ! git diff --name-only --exit-code HEAD^1 HEAD CHANGELOG.md
\ No newline at end of file
+        run: "! git diff --name-only --exit-code HEAD^1 HEAD CHANGELOG.md"
```
bcardarella, line 18: I'm not sure this is needed as not every PR will affect public API

carson-katri: We can make it so that it doesn't block merging, but still works as a reminder to update the changelog.

PR #1487, opened 2024-11-15
Add `:section` slot to `simple_form`.

Serabe:
This PR adds a `:section` slot to `simple_form/1`. The slot can receive the following optional attributes:

- ~`is_expanded`, boolean. Will be set as `isExpanded` in the sections. Defaults to true.~
- `header`, string. Will add a `<Text template={:header}></Text>` with the same content to the `Section`.
- `footer`, string. Same as header, but with template `footer`.

Also renders conditionally the `Section` containing the actions.

One final caveat is that if any `:section` slot is present, then `@inner_block` is not rendered. This prevents undefined behaviour or expectations to arise.

## Questions

1. ~Should `is_expanded` be `isExpanded` to be closer to SwiftUI or keep it closer to Elixir?~
2. I could not figure out how to read the SwiftUI documentation so I'm not sure if I'm missing something.

### DIFF
--- a/priv/templates/lvn.swiftui.gen/core_components.ex
+++ b/priv/templates/lvn.swiftui.gen/core_components.ex
@@ -416,14 +416,11 @@ defmodule <%= inspect context.web_module %>.CoreComponents.<%= inspect context.m
           <%%= render_slot(@inner_block, f) %>
         <%% else %>
           <%%= for section <- @section do %>
-            <Section isExpanded={Map.get(section, :is_expanded, true)}>
-              <Text :if={not is_nil(Map.get(section, :header))} template={:header}>
-                <%%= Map.get(section, :header) %>
-              </Text>
+            <Section>
+              <Text :if={not is_nil(Map.get(section, :header))} template={:header} content={Map.get(section, :header)} />
+
               <%%= render_slot(section) %>
-              <Text :if={not is_nil(Map.get(section, :footer))} template={:footer}>
-                <%%= Map.get(section, :footer) %>
-              </Text>
+              <Text :if={not is_nil(Map.get(section, :footer))} template={:footer} content={Map.get(section, :footer)} />
             </Section>
           <%% end %>
         <%% end %>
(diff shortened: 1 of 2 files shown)


bcardarella:
This lgtm but I'll want @carson-katri to review before merge

priv/templates/lvn.swiftui.gen/core_components.ex
```diff
@@ -400,15 +400,34 @@ defmodule <%= inspect context.web_module %>.CoreComponents.<%= inspect context.m
...
-        <%%= render_slot(@inner_block, f) %>
-        <Section>
+        <%%= if @section == [] do %>
+          <%%= render_slot(@inner_block, f) %>
+        <%% else %>
+          <%%= for section <- @section do %>
+            <Section isExpanded={Map.get(section, :is_expanded, true)}>
+              <Text :if={not is_nil(Map.get(section, :header))} template={:header}>
+                <%%= Map.get(section, :header) %>
+              </Text>
```
carson-katri, line 422: You could use an attribute here, since you're just passing a string value into the `Text`:
```suggestion
              <Text :if={not is_nil(Map.get(section, :header))} template={:header} content={Map.get(section, :header)} />
```

test/live_view_native/swiftui/core_components_test.exs
```diff
@@ -340,7 +344,176 @@ defmodule LiveViewNative.SwiftUI.CoreComponentsTest do
...
+              <VStack alignment="leading">
+                <TextField id="user_email" name="user[email]" style="" text="test@example.com">Email</TextField>
+              </VStack>
+            </Section>
+          </Form>
+        </LiveForm>
+        """)
+    end
+
+    test "can render sectionn with footer" do
```
carson-katri, line 432:
```suggestion
    test "can render section with footer" do
```

carson-katri (approved):
I think we should follow Elixir naming conventions in a component, so `is_expanded` works better.

bcardarella:
@Serabe please remove the formatting changes. This PR includes changes to functions that are outside of its scope

bcardarella:
yeah I'll correct that in another PR. I'll reivew this in a bit thanks!

PR #1497, opened 2024-12-13
Add `isExpanded` to Section to replace `collapsible`

carson-katri:
If `isExpanded` is not provided, the section cannot be collapsed/expanded. If provided, it expects a Bool.


bcardarella:
Can you add to CHANGELOG

PR #1503, opened 2024-12-22
rc-4 changes: Add additional headers to connect, reconnect endpoints on coordinator

mobile-bungalow:
We need the ability to add additional headers in `connect`, `reconnect`. phoenix will throw a csrf error on live form submission unless `content-type` is set to `application/x-www-form-urlencoded`.  This can already be done by changing the configured headers on the coordinator, but this is more explicit and avoid bugs where setting and unsetting them globally leaves them in the wrong state after a failed connection.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,13 +1,13 @@
 {
-  "originHash" : "6a5ddfa26364abe0acd53c4af877d94a75bd5865a0fe9de361261256db6916f3",
+  "originHash" : "f9084835ddc2d4db4a43a936bfe07b6b768f92c32fab7a578eb25826ad91bd16",
   "pins" : [
     {
       "identity" : "liveview-native-core",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/liveview-native/liveview-native-core",
       "state" : {
-        "revision" : "74f5044cba915d87cbc4b882b23a1bc6291b3a37",
-        "version" : "0.4.0-rc-3"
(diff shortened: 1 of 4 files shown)


bcardarella:
@mobile-bungalow are those needs in addition to this PR or part of this PR?

mobile-bungalow:
They are enabled by this PR, the changes need to happen in [liveview-native-liveform](https://github.com/liveview-native/liveview-native-live-form/blob/33bce0879fa0604113cb22109795f183ff0f36e4/swiftui/Sources/LiveViewNativeLiveForm/LiveForm.swift#L62-L67) upon submitting the form.

bcardarella:
That doesn't sound correct to me. `application/x-www-form-urlencoded` is the older `contet-type`, `multipart/form-data` is the modern one. For example, the Fetch API in JavaScript will default to `multipart/form-data` for all `POST` requests. I think we should just have LVN Core default `POST` requests to `multipart/form-data` too but allow for an optional override of the `content-type` from the client if desired. However it is highly unlikely there will be a usecase for the downgraded `content-type`.

In otherwords, the clients should get something similar to the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) that they can call from Core.

bcardarella:
Actually, I'm incorrect. The Fetch API will only default to a `multipart/form-data` if the `body` is a `FormData` object.

PR #1509, opened 2025-01-02
Properly decode "e" payload for handling events

carson-katri:
The `"e"` key was being found, but the individual events were not being properly extracted from core's `Json` types.

Now the string name and JSON object and fetched properly.

I've also added a `Decodable` API for `receiveEvent`:

```swift
struct MyEvent: Decodable {
    let message: String
}

struct ContentView: View {
    @StateObject private var session = LiveSessionCoordinator(.localhost)

    var body: some View {
        LiveView(...) { ... }
            .onReceive(session.receiveEvent("test", for: MyEvent.self), perform: { (coordinator, event) in
                print(event) // MyEvent(message: "hello world")
            })
    }
}
```

```ex
{:noreply, socket |> push_event("test", %{ "message" => "hello world" })}
```


carson-katri:
I've added a default value to the `for` argument in `receiveEvent`. So you can now have the Swift compiler infer the argument from an annotated closure if you prefer:

```swift
.onReceive(session.receiveEvent("test")) { (coordinator: _, event: MyEvent) in
    ...
}
```

nighthawk:
Just spotted this PR. I can confirm that this fixes #1526. Thank you @carson-katri.

PR #1511, opened 2025-01-04
Use previous channel in navigation API

mobile-bungalow:
The previous channel must be used in the navigation API to properly close the channel and reuse the connection parameters.


mobile-bungalow:
upon further testing, if the previous navigator failed to open a`LiveChannel` this fails. this will be irrelevant when the refactor is complete, however.

Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -107,20 +107,21 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
         $navigationPath.scan(([LiveNavigationEntry<R>](), [LiveNavigationEntry<R>]()), { ($0.1, $1) }).sink { [weak self] prev, next in
             guard let self else { return }
             Task {
+                let last_channel = prev.last?.coordinator.liveChannel!
```
carson-katri, line 110: This should be `lastChannel` in camelCase

mobile-bungalow: my poor rust addled brain.

PR #1518, opened 2025-02-05
Post Refactor update to session coordinator

mobile-bungalow:
here are speculative changes that would need to be made to the swiftui client if [my refactor](https://github.com/liveview-native/liveview-native-core/pull/289) landed in it's current state.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
@@ -107,20 +107,21 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
         $navigationPath.scan(([LiveNavigationEntry<R>](), [LiveNavigationEntry<R>]()), { ($0.1, $1) }).sink { [weak self] prev, next in
             guard let self else { return }
             Task {
+                let last_channel = prev.last?.coordinator.liveChannel!
                 try await prev.last?.coordinator.disconnect()
                 if prev.count > next.count {
                     // back navigation (we could be going back multiple pages at once, so use `traverseTo` instead of `back`)
                     let targetEntry = self.liveSocket!.getEntries()[next.count - 1]
                     next.last?.coordinator.join(
-                        try await self.liveSocket!.traverseTo(targetEntry.id, next.last!.coordinator.liveChannel, nil)
+                        try await self.liveSocket!.traverseTo(targetEntry.id, last_channel, nil)
                     )
                 } else if next.count > prev.count && prev.count > 0 {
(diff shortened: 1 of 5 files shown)


Sources/LiveViewNative/Coordinators/LiveSessionCoordinator.swift
```diff
@@ -103,26 +128,40 @@ public class LiveSessionCoordinator<R: RootRegistry>: ObservableObject {
...
+                    let _ = try await client.navigate(next.last!.url.absoluteString, opts)
                 } else if next.count == prev.count {
-                    guard let liveChannel = try await self.liveSocket?.navigate(next.last!.url.absoluteString, next.last!.coordinator.liveChannel, NavOptions(action: .replace))
-                    else { return }
-                    next.last?.coordinator.join(liveChannel)
+                    // TODO: this will fire on a patch event! this should not fire on a patch event
+                    var opts = NavOptions()
+                    opts.joinParams = .some([ "_interface": .object(object: LiveSessionParameters.platformParams)])
+                    opts.action = .replace
+                    let _ = try await client.navigate(next.last!.url.absoluteString, opts)
```
mobile-bungalow, line 164: There is a problem here, core now runs it's own even loop and will take patch and redirect events, these are currently not handled if they are not triggered by a client action. Those events are all pushed through `SimpleEventHandler.viewReloadSubject`. We need to watch for events that are emitted by server-side events and rectify the navigation path. 

the other option is just to call the appropriate client navigation functions wherever we would previously just modify the navigation path.

bcardarella: I'm not following the specific issue here. Can you provide more detail?

mobile-bungalow: previously the swiftui client listened on the live view channel for `live_reload` and other `push_navigation` type events, then it would modify the `LiveSessionCoordinator.NavigationPath` and all navigation would happen downstream of checking for changes in the `scan` block like we do here. 

Now that push navigation events are handled opaquely inside core, we need to update the `LiveSessionCoordinator.NavigationPath` in response to internal events. While the code provided here works, the `NavigationPath` should be controlled *by*  core, which is the opposite of how it works now.

bcardarella: oh nice :)

bcardarella: and yes, I agree

mobile-bungalow:
Hand merged changes into #1549

PR #1528, opened 2025-02-12
New member expression syntax

bcardarella:
Per-Carson's request the representation of member expressions is going to change to make the SwiftUI parsing easier to reason about

@NduatiK I've made a few updates to the tsets but after those tests are green I suspect other tests will fail as the new format hasn't been update everywhere in the test suite,

More details on the syntax

Screenshot 2025-02-11 at 10 56 27 PM

I seem to recall that @NduatiK or @carson-katri recommended this format last year.

### DIFF
--- a/test/live_view_native/swiftui/rules_parser_test.exs
+++ b/test/live_view_native/swiftui/rules_parser_test.exs
@@ -96,8 +96,8 @@ defmodule LiveViewNative.SwiftUI.RulesParserTest do
 
       assert parse(input) == output
 
-      input = "Color.blue.opacity(0.2).blendMode(.multiply).mix(with: .orange).opacity(0.7)"
-      output = {:., [], [{:., [], [{:., [], [{:., [], [{:., [], [:Color, :blue]}, {:opacity, [], [0.2]}]}, {:blendMode, [], [{:., [], [nil, :multiply]}]}]}, {:mix, [], [[with: {:., [], [nil, :orange]}]]}]}, {:opacity, [], [0.7]}]}
+      input = "background(Color.blue.opacity(0.2).blendMode(.multiply).mix(with: .orange).opacity(0.7))"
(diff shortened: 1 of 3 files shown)


carson-katri:
Member access is essentially inverted.

Currently if you access a member on a static member, it will encode them as the left and right of a 'dot' (.) operation with a . in front of both of them. I think structurally it makes more sense for the static member to be evaluated first, and the other members to be separate 'dot' operations that wrap it.

```javascript
// usage
.staticMember.instanceMember

// current format:
[".", {}, [null, [".", {}, ["staticMember", "instanceMember"]]]]
// interpret as `.(staticMember.instanceMember)`

// proposed format:
[".", {}, [[".", {}, [null, "staticMember"]], "instanceMember"]]
// interpret as `(.staticMember).instanceMember`
```

with this new format we can more easily compose decoders to handle the AST:

```
(base decoder) = .staticMember
(member decoder) = (base decoder).instanceMember
```

bcardarella:
@NduatiK I think I got it solved

test/live_view_native/swiftui/rules_parser_test.exs
```diff
@@ -179,15 +182,7 @@ defmodule LiveViewNative.SwiftUI.RulesParserTest do
...
-      output =
-        {:color, [],
-         [
-           {
-             :color,
-               {:., [],
-                [nil, {:., [], [:foo, {:., [], [:bar, {:., [], [{:baz, [], [1, 2]}, :qux]}]}]}]}
-           }
-         ]}
+      output = {:color, [], [color: {:., [], [{:., [], [{:., [], [{:., [], [nil, :foo]}, :bar]}, {:baz, [], [1, 2]}]}, :qux]}]}
```
bcardarella, line 185: I think this is wrong, I'm looking into it....

bcardarella: I believe this passes simply because elixir lists such as `[foo: 123]` is really `[{:foo, 123}]` but it should work for us either way

PR #1532, opened 2025-02-14
retain cycle fix

mobile-bungalow:
because the channel never is upgraded to a strong reference it is retained after close, you can confirm this in the xcode memory graph view.

### DIFF
--- a/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
+++ b/Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
@@ -239,8 +239,7 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
 
     func bindEventListener() {
         self.eventListenerLoop = Task { [weak self, weak channel] in
-            guard let channel else { return }
-            let eventListener = channel.eventStream()
+            let eventListener = channel!.eventStream()
             for try await event in eventListener {
                 guard let self else { return }
                 guard !Task.isCancelled else { return }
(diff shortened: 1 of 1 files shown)


Sources/LiveViewNative/Coordinators/LiveViewCoordinator.swift
```diff
@@ -239,8 +239,7 @@ public class LiveViewCoordinator<R: RootRegistry>: ObservableObject {
 
     func bindEventListener() {
         self.eventListenerLoop = Task { [weak self, weak channel] in
-            guard let channel else { return }
-            let eventListener = channel.eventStream()
+            let eventListener = channel!.eventStream()
```
carson-katri, line 242: Try changing `weak channel` to `unowned channel`, it should have the same effect as force unwrapping the weak channel.

mobile-bungalow: I just tried: 

```swift
self.eventListenerLoop = Task { [weak self, unowned channel] in
            guard let channel else { return }
            let eventListener = channel.eventStream()
```

In the graph view this creates an owning reference to the channel that is retained.

should I commit

```swift
self.eventListenerLoop = Task { [weak self, unowned channel] in
            let eventListener = channel!.eventStream()
```

instead?

carson-katri: Yeah remove the guard statement

PR #1535, opened 2025-02-19
Upgrade core to 0.4.1-rc1

mobile-bungalow:
Additionally: cleanup memory leaks which leak web socket connections. This means we will postpone moving to the LiveViewClient API until 0.4.1

### DIFF
--- a/Sources/LiveViewNative/Live/LiveView.swift
+++ b/Sources/LiveViewNative/Live/LiveView.swift
@@ -76,6 +76,11 @@ public macro LiveView<
     @ViewBuilder content: @escaping (LiveViewPhase<EmptyRegistry>) -> PhaseView = { (_: LiveViewPhase<EmptyRegistry>) -> Never in fatalError() }
 ) -> AnyView = #externalMacro(module: "LiveViewNativeMacros", type: "LiveViewMacro")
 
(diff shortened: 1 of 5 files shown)


Sources/LiveViewNative/Live/LiveView.swift
```diff
@@ -245,17 +241,21 @@ public struct LiveView<
...
+        .environment(\.eventConfirmation, session.configuration.eventConfirmation ?? { [weak session] message, _ in
             return await withCheckedContinuation { continuation in
-                eventConfirmationTransaction = EventConfirmationTransaction(message: message, callback: continuation.resume(returning:))
-                showEventConfirmation = true
+                if let session = session {
+                    session.eventConfirmationTransaction = EventConfirmationTransaction(message: message, callback: continuation.resume(returning:))
+                    session.showEventConfirmation = true
+                } else {
+                    continuation.resume(returning: false)
+                }
```
mobile-bungalow, line 251: This is a particularly insidious retain cycle I want to draw attention to.  `self` was captured in a way that was retained, which captured `session` with it.  There are other leaks like this which we ought to track down but that have less immediate consequences.

carson-katri (approved):

